PR #9, opened 2018-03-08
ignore targets and doc tweaks

michaelneale:


jonhermansen (commented):
👍

jonhermansen:
@michaelneale no trailing newline? 😢

michaelneale:
@jonhermansen fixed - also added some corrections about the path to the plugins dir (and I pointed it to the LTS vs weekly...)

PR #27, opened 2018-05-31
Optimize docker build process by caching maven dependencies

toshke:
Retesting docker build process took me a while if there is an syntax error in the code or such. This approach caches all of maven dependencies in separate layer, thus shortening time to repetitively build container. Also this PR updates `Dockerfile` to use -x style parameters implemented in https://github.com/kohsuke/jenkinsfile-runner/pull/21.

### DIFF
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -8,15 +8,26 @@ Edit the `plugins.txt` to to include any plugins you'd like to install. See [the
 Then, build the Jenkinsfile Runner image like this:
 
 ```
-docker build -t jenkinsfile-runner:my-production-jenkins --build-arg JENKINS_VERSION=2.108 .
+docker build -t jenkinsfile-runner:my-production-jenkins --build-arg JENKINS_VERSION=2.121.1 .
 ```
+
+If you are rebuilding against latest `lts` image, pass `--no-cache` argument to command above
+to avoid using stale layers
 
 The optional `JENKINS_VERSION` specifies the version of Jenkins core.
 
 ## Usage
-Run the image by mounting the directory that contains `Jenkinsfile` into `/workspace`. For example,
+Run the image by providing path to `Jenkinsfile` . For example,
 
-```
-docker run -v~/foo:/workspace jenkinsfile-runner:my-production-jenkins
+```bash
+docker run --rm -v $PWD/test:/workspace jenkinsfile-runner:my-production-jenkins
 ```
 
+Optionally, if you want to change default parameters for plugins or workspace, you can get onto the container
+by overriding entrypoint - binary is placed in `/app/bin/jenkinsfile-runner`
+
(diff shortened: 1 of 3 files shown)


schnatterer:
Hey @toshke , caching the maven deps speeds up recurring builds so much. Nice idea.

I don't understand why you changed the usage of the docker container, though.

* It was:   
  `docker run -v~/foo:/workspace jenkinsfile-runner:my-production-jenkins`
* now would be:
  `docker run -v~/foo:/workspace -w /workspace jenkinsfile-runner:my-production-jenkins -f test/Jenkinsfile `

I don't want to judge which one is better. Changing it, forces all existing user to adopt, though. So there should be a good reason.

With the following entrypoint, would could keep the usage of the container the same by adopting the the new `-x` style parameters from #21.
`ENTRYPOINT ["/app/bin/jenkinsfile-runner", "-w", "/app/jenkins", "-p", "/usr/share/jenkins/ref/plugins", "-f", "/workspace"]`

In the `DOCKER.md` you describe that the default parameters for plugins or workspace can be overwritten, which is correct. The example bellow then only sets the entrypoint with `bash`. I don't get the point here. Wouldn't an example that mounts a plugin folder and overwrite the entrypoint with `-p`  fit better?

toshke:
@schnatterer glad to be helpful, this PR actually is side-effect of work on https://github.com/kohsuke/jenkinsfile-runner/pull/26

As for update usage of container - you are right, I haven't really take backward compatibility into consideration - i've amended commit to include your proposed entrypoint.

I've also set default jenkins version to `2.121.1` as pipeline plugin wouldn't load with previous versions. One suggestion would be to use `lts` (though this requires build without caching everytime upstream image updates).. For some reason plugin installation get stuck when using `workflow-aggregator` as default plugin, so i've update this to use `pipeline-model-definition` plugin (and tested with success). 

Furthermore, as for documentation wise, my not-so-well explained point was that binary is available to be consumed by logging into container as well. I've updated doco to better demonstrate this possibility (below, and in amended commit)

--------
## Usage
Run the image by providing path to `Jenkinsfile` . For example,

```bash
docker run --rm -v $PWD/test:/workspace  jenkinsfile-runner:my-production-jenkins
```

Optionally, if you want to change default parameters for plugins or workspace, you can get onto the container
by overriding entrypoint - binary is placed in `/app/bin/jenkinsfile-runner`

```bash
$ docker run --rm -it -v $PWD/test:/workspace --entrypoint bash jenkinsfile-runner:my-production-jenkins
root@dec4c0f12478:/src# cp -r /app/jenkins /tmp/jenkins
root@dec4c0f12478:/src# /app/bin/jenkinsfile-runner -w /tmp/jenkins -p /usr/share/jenkins/ref/plugins -f /workspace
```

PR #29, opened 2018-09-06
Start http server if port property is set

carlossg:
This may be useful for something, like connecting jnlp agents


oleg-nenashev:
@carlossg no strong opinion.  Is it still needed if we get https://github.com/jenkinsci/remoting/pull/286 merged/released?

carlossg:
not needed

PR #32, opened 2018-10-30
Create a runnable Jenkinsfile Runner demo.

oleg-nenashev:
This is a copy of a more complicated CWP demo, but I believe it’s useful to have something in this repo as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ bootstrap/target/
 payload-dependencies/target/
 payload/target/
 setup/target/
+cwp-cli-1.3.jar
--- a/demo/cwp/Makefile
+++ b/demo/cwp/Makefile
@@ -1,0 +1,24 @@
+# Just a Makefile for manual testing
+.PHONY: all
+
+ARTIFACT_ID = jenkinsfile-runner-demo
+VERSION = 256.0-test
+CWP_VERSION = 1.3
+
+all: clean build
+
+clean:
+	rm -rf tmp
+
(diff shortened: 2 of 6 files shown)


jvz (approved):

demo/cwp/Makefile
```diff
@@ -0,0 +1,24 @@
+# Just a Makefile for manual testing
+.PHONY: all
```
alecharp, line 2: `clean` and `run` should also be part of `PHONY`.

demo/cwp/Makefile
```diff
@@ -0,0 +1,24 @@
...
+ARTIFACT_ID = jenkinsfile-runner-demo
+VERSION = 256.0-test
+CWP_VERSION = 1.3
+
+all: clean build
+
+clean:
+	rm -rf tmp
+
+.build/cwp-cli-${CWP_VERSION}.jar:
```
alecharp, line 13: why did you prefixed it with a `.`? If this is a real file, then don't, but add `build` in `PHONY` listing.

.gitignore
```diff
@@ -3,3 +3,4 @@ bootstrap/target/
 payload-dependencies/target/
 payload/target/
 setup/target/
+cwp-cli-1.3.jar
```
alecharp, line 6: From the `Makefile`, shouldn't this file be in `build`?

oleg-nenashev: Yes, historical stuff

PR #33, opened 2018-10-30
Update Jenkins Test Harness to the recent version with Patches and Java 11 support.

oleg-nenashev:
Work on my machine (c).


oleg-nenashev:
@ndeloof @schnatterer Any concerns if I just start merging such stuff without review as long as local tests pass?

ndeloof:
Works for me

jvz (approved):

schnatterer:
TBH I never had too much insight into the Java code, so can't judge if the test coverage is sufficient. Unfortunately, still no free time to work on that insight. 

So, no objections from my side 😉

PR #34, opened 2018-10-30
Fix Maven Enforcer in the codebase

oleg-nenashev:
The change fixes the reported Enforcer issues... and it kinda works in the reality. In the default bundle there are some issues with symbol resolution


oleg-nenashev:
Would also blow up with Java 11, so I will merge it for now

PR #35, opened 2018-10-30
[JENKINS-54485] - Add integration tests based on the CWP demo

oleg-nenashev:
[JENKINS-54485](https://issues.jenkins-ci.org/browse/JENKINS-54485)

CC @jvz

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -64,7 +64,7 @@ demos['cwp'] = {
             checkout scm
             stage('CWP') {
                 dir('demo/cwp') {
-                    sh "make clean build run"
+                    sh "make clean buildInDocker run"
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


varyvol:
@oleg-nenashev I have pushed some commits to your branch:
- The Docker image is built outside to avoid using docker in docker.
- The source needs to be copied so that the build works both in the classic way and with docker.

Feel free to review!

oleg-nenashev (commented):
@varyvol  I am not sure Make is going to track dependencies correctly after introducing `.copySource`. Have you verified that the `make build` does not not crash on second time when `clean` is not invoked? If it crashes, it may make sense to make `clean` a strong requirement

Regarding the rest, it looks good to me. Thanks @varyvol !

varyvol:
@oleg-nenashev yes, it works fine, `rsync` checks the `source` folder is already there and returns successfully.

PR #36, opened 2018-11-02
[JENKINS-54424] - Remove Jenkins Test Harness Dependency & replace by local code

oleg-nenashev:
This change removes dependency on the Jenkins Test Harness framework. JUnit is still used, but there is a separate ticket for that.

* Many classes were copied from JTH, but it can be optimized later
* Irrelevant code in JenkinsRule and other classes was removed

There is still a lot of optimizations/refactorings to be done (e.g. prevent WAR exploder from running at all, remove interfaces), but at least we keep the engine separate from JTH going forward

https://issues.jenkins-ci.org/browse/JENKINS-54424


setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsfileRunnerRule.java
```diff
@@ -47,13 +42,8 @@ public JenkinsfileRunnerRule(File warDir, File pluginsDir) {
...
     @Override
     protected ServletContext createWebServer() throws Exception {
-        server = new Server(new ThreadPoolImpl(new ThreadPoolExecutor(10, 10, 10L, TimeUnit.SECONDS, new LinkedBlockingQueue<Runnable>(),new ThreadFactory() {
-            public Thread newThread(Runnable r) {
-                Thread t = new Thread(r);
-                t.setName("Jetty Thread Pool");
-                return t;
-            }
-        })));
+        QueuedThreadPool queuedThreadPool = new QueuedThreadPool(10);
```
oleg-nenashev, line 45: Removed custom Thread Pool like @olamy did for Winstone in https://github.com/jenkinsci/winstone/pull/54 . Likely we need to upstream the fix to Jenkins Test Harness so that it runs reliably on 50+ core machines. I wish we had them on our infra :P 

FYI @jglick

setup/src/main/java/io/jenkins/jenkinsfile/runner/util/HudsonHomeLoader.java
```diff
@@ -0,0 +1,165 @@
...
+     *
+     * @throws Exception
+     *      to cause a test to fail.
+     */
+    File allocate() throws Exception;
+
+    /**
+     * Allocates a new empty directory, meaning this will emulate the fresh Hudson installation.
+     */
+    HudsonHomeLoader NEW = new HudsonHomeLoader() {
```
jvz, line 55: Can't this be a lambda function?

setup/src/main/java/io/jenkins/jenkinsfile/runner/util/HudsonHomeLoader.java
```diff
@@ -0,0 +1,165 @@
...
+            this.source = source;
+        }
+
+        public File allocate() throws Exception {
+            File target = NEW.allocate();
+            if(source.getProtocol().equals("file")) {
+                File src = new File(source.toURI());
+                if(src.isDirectory())
+                    new FilePath(src).copyRecursiveTo("**/*",new FilePath(target));
+                else
```
jvz, line 91: Might want to add some curly braces here for clarity.

setup/src/main/java/io/jenkins/jenkinsfile/runner/util/JenkinsRecipe.java
```diff
@@ -0,0 +1,57 @@
...
+     */
+    Class<? extends Runner> value();
+
+    /**
+     * The code that implements the recipe semantics.
+     *
+     * @param <T>
+     *      The recipe annotation associated with this runner.
+     */
+    abstract class Runner<T extends Annotation> {
```
jvz, line 40: Not sure about compatibility, but would it be possible to refactor this into an interface with default methods?

setup/src/main/java/io/jenkins/jenkinsfile/runner/util/JenkinsRecipe.java
```diff
@@ -0,0 +1,57 @@
+package io.jenkins.jenkinsfile.runner.util;
```
jvz, line 1: Needs a license header.

setup/src/main/java/io/jenkins/jenkinsfile/runner/util/EndOfTestListener.java
```diff
@@ -0,0 +1,16 @@
+package io.jenkins.jenkinsfile.runner.util;
```
jvz: Needs a license header.

setup/src/main/java/io/jenkins/jenkinsfile/runner/util/TemporaryDirectoryAllocator.java
```diff
@@ -0,0 +1,110 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Allocates temporary directories and cleans it up at the end.
+ * @author Kohsuke Kawaguchi
+ */
+public class TemporaryDirectoryAllocator {
```
jvz, line 37: Might be too late at this point, but doesn't the JUnit TemporaryFolder rule do this, too?

jvz (changes requested):
Assuming you just copied most of this code, the only real fixes I would need are the copyright headers.

setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsRule.java
```diff
@@ -0,0 +1,1276 @@
...
+    }
+
+    /**
+     * Backward compatibility with JUnit 4.8.
+     */
+    public Statement apply(Statement base, FrameworkMethod method, Object target) {
+        return apply(base,Description.createTestDescription(method.getMethod().getDeclaringClass(), method.getName(), method.getAnnotations()));
+    }
+
+    public Statement apply(final Statement base, final Description description) {
```
jglick: Why is this here??

setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsRule.java
```diff
@@ -0,0 +1,1276 @@
...
+    /**
+     * Prepares a webapp hosting environment to get {@link javax.servlet.ServletContext} implementation
+     * that we need for testing.
+     */
+    protected abstract ServletContext createWebServer() throws Exception;
+
+    /**
+     * Configures a security realm for a test.
+     */
+    protected LoginService configureUserRealm() {
```
jglick: delete

setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsRule.java
```diff
@@ -0,0 +1,1276 @@
...
+        try {
+            return new Hudson(home, webServer, getPluginManager());
+        } catch (InterruptedException x) {
+            throw new AssumptionViolatedException("Jenkins startup interrupted", x);
+        } finally {
+            jettyLevel(Level.INFO);
+        }
+    }
+
+    public PluginManager getPluginManager() {
```
jglick: delete

setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsRule.java
```diff
@@ -0,0 +1,1276 @@
...
+    public <T extends TopLevelItem> T createProject(Class<T> type, String name) throws IOException {
+        return jenkins.createProject(type, name);
+    }
+
+    /**
+     * Creates a new job with an unique name.
+     *
+     * @param type Top level item type.
+     */
+    public <T extends TopLevelItem> T createProject(Class<T> type) throws IOException {
```
jglick: delete

jglick (commented):
There is a ton of stuff here that does not look necessary at all. I do not understand why we would bother to remove the JTH binary dependency, only to copy & paste all this stuff which only exists for tests and has no relevance to the purpose.

Would it not be simpler to throw out this repo and start with @ndeloof’s, which seems to have a cleaner design?

oleg-nenashev:
> There is a ton of stuff here that does not look necessary at all. I do not understand why we would bother to remove the JTH binary dependency, only to copy & paste all this stuff which only exists for tests and has no relevance to the purpose.

The code will change late to improve performance. By moving the code here, we have no risk of breaking the JTH use-cases

> Would it not be simpler to throw out this repo and start with @ndeloof’s, which seems to have a cleaner design?

Performance, again. This repo, even it has a non-production design (powered by JUnit, yey!), gave me much better startup time than the one from @ndeloof . Assuming I find enough time, I want to gradually integrate Nicolas' codebase and to move performance tweaks to the Jenkins core directly where possible

setup/src/main/java/io/jenkins/jenkinsfile/runner/util/TemporaryDirectoryAllocator.java
oleg-nenashev: My plan is to get rid of JUnit... and this implementation as well. We really need a really simple mapper to a fixed directory, no magic required for Jenkinsfile Runner

setup/src/main/java/io/jenkins/jenkinsfile/runner/util/JenkinsRecipe.java
oleg-nenashev: Would prefer to create a follow-up task for now

PR #41, opened 2018-11-30
[NO_MERGE] [JENKINS-54905] - Experimental branch for Java 11 support

oleg-nenashev:
Currently it crashes, but I am staging what I was able to achieve.

- [x] [JENKINS-54905](https://issues.jenkins-ci.org/browse/JENKINS-54905) - Update Docker image for Java 11
- [x] Lame attempt to fix [JENKINS-54426](https://issues.jenkins-ci.org/browse/JENKINS-54426) via Join classloader, didn't work

@jenkinsci/java11-support @varyvol @fcojfernandez


demo/cwp/packager-config.yml
```diff
@@ -14,14 +14,14 @@ buildSettings:
       source:
         dir: ../..
     docker:
-      base: "jenkins/jenkins:2.138.2"
+      base: "jenkins/jenkins-experimental:latest-jdk11"
```
batmat, line 17: Don't you want to use `jenkins/jenkins` now we got the official JDK 11 support there, like you did above?

Dockerfile
```diff
@@ -1,7 +1,7 @@
-ARG JENKINS_VERSION=2.121.1
+ARG JENKINS_VERSION=jdk11
```
batmat, line 1: Do you really intend to modify JFR's master branch with using JDK11 as default? Or maybe we don't care because packagers are always expected to pass this `ARG`?

plugins.txt
```diff
@@ -1 +1,2 @@
 pipeline-model-definition:latest
+workflow-support:incrementals;org.jenkins-ci.plugins.workflow;2.21-rc591.43d37d4d080a
```
batmat, line 2: Might make sense to bump to `2.23-rc674.73bb98187744` for clarity and consistency with the one we ended up releasing and use on Evergreen: https://github.com/jenkins-infra/evergreen/blob/36a60c268be9529c0f5c019b9ac1b0f192b3032a/services/essentials.yaml#L171

setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsfileRunnerRule.java
```diff
@@ -46,7 +54,16 @@ protected ServletContext createWebServer() throws Exception {
         server = new Server(queuedThreadPool);
 
         WebAppContext context = new WebAppContext(warDir.getPath(), contextPath);
-        context.setClassLoader(getClass().getClassLoader());
+        String javaVersion = System.getProperty("java.version");
+
+//        if (javaVersion.startsWith("1.")) {
```
batmat, line 59: Let's remove the commented out code?

Also, if you mean to keep it, then I guess you'd use `isPostJava8()`

setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsfileRunnerRule.java
```diff
@@ -46,7 +54,16 @@ protected ServletContext createWebServer() throws Exception {
         server = new Server(queuedThreadPool);
 
         WebAppContext context = new WebAppContext(warDir.getPath(), contextPath);
-        context.setClassLoader(getClass().getClassLoader());
+        String javaVersion = System.getProperty("java.version");
```
batmat, line 57: unused, I guess leftover from the new `isPostJava8()` method introduced above.

setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsfileRunnerRule.java
```diff
@@ -46,7 +54,16 @@ protected ServletContext createWebServer() throws Exception {
...
+
+//        if (javaVersion.startsWith("1.")) {
+//            context.setClassLoader(getClass().getClassLoader());
+//        } else {
+            ClassLoader platform = (ClassLoader) ClassLoader.class.getMethod("getPlatformClassLoader").invoke(null);
```
batmat, line 62: Can we have a comment explaining the intent a bit here?

demo/cwp/packager-config.yml
oleg-nenashev: agreed

setup/src/main/java/io/jenkins/jenkinsfile/runner/util/JoinClassLoader.java
```diff
@@ -0,0 +1,90 @@
+package io.jenkins.jenkinsfile.runner.util;// Copyright 2010 Christian d'Heureuse, Inventec Informatik AG, Zurich, Switzerland
+// www.source-code.biz, www.inventec.ch/chdh
+//
+// This module is multi-licensed and may be used under the terms
+// of any of the following licenses:
+//
+//  LGPL, GNU Lesser General Public License, V2 or later, http://www.gnu.org/licenses/lgpl.html
```
oleg-nenashev, line 7: Note that this code is not compatible w.r.t the license. I put it here only for the experiment (which does not work so far), and it should not be merged as is

batmat:
This is now conflicted FTR.

setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsfileRunnerRule.java
varyvol: Yes please!

fcojfernandez: IIUC, the goal here is get exactly the same classloader as in previous JDK versions. Since this piece of code is executed always and not only with java11, what will happen if this code is executed with java8? The `getPlatformClassLoader` is only available from java9 (if I'm not wrong), so it will throw an exception

varyvol:
@oleg-nenashev how do you want to proceed with this PR?

oleg-nenashev:
@varyvol I think the best way is to...

1) Backup the branch in my local fork
2) Delete the `jdk11` branch from this repo so that we can reuse it for actual changes

oleg-nenashev:
Moved the code to https://github.com/oleg-nenashev/jenkinsfile-runner/tree/jdk11-experimental . After patches by @varyvol I doubt these patches will be actually needed

PR #42, opened 2018-12-07
[JENKINS-54351] Avoid skipping enforcer in builds

varyvol:
[JENKINS-54351](https://issues.jenkins-ci.org/browse/JENKINS-54351)

Avoid skipping the enforcer in builds. Exclude a couple dependencies that were raising upper bound errors.

@oleg-nenashev @fcojfernandez

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ payload-dependencies/target/
 payload/target/
 setup/target/
 cwp-cli-1.3.jar
+demo/cwp/tmp/output
+demo/cwp/tmp/prebuild
--- a/bootstrap/pom.xml
+++ b/bootstrap/pom.xml
@@ -14,12 +14,12 @@
     <dependency>
       <groupId>args4j</groupId>
       <artifactId>args4j</artifactId>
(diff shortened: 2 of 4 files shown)


varyvol:
@oleg-nenashev I executed the CWP demo and everybody seems fine. Any other test that you think could be useful to execute here?

fcojfernandez (approved):

oleg-nenashev (changes requested):
De-facto versions from the core will be used in Jenkins. So it rather hides the problem than solves it.

varyvol:
@oleg-nenashev I avoided exclusions by reducing the versions, but I don't have context on why those were used.

oleg-nenashev (approved):
AFAICT there is nothing specific in args4j which would be needed here: https://github.com/kohsuke/args4j/compare/args4j-site-2.0.31...args4j-site-2.33

Would make sense to bump Jenkins, but no immediate need to do that

oleg-nenashev:
Thanks @varyvol !

PR #49, opened 2018-12-20
[JENKINS-54425] Avoid warnings in build logs

varyvol:
In general we don't want warnings in build logs, since we are only interested there in the results of our Jenkinsfiles.

**BEFORE**
screen shot 2018-12-20 at 09 38 05

**AFTER**
screen shot 2018-12-20 at 11 37 01

### DIFF
--- a/setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsfileRunnerRule.java
+++ b/setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsfileRunnerRule.java
@@ -1,5 +1,6 @@
 package io.jenkins.jenkinsfile.runner;
 
+import hudson.ClassicPluginStrategy;
 import hudson.security.ACL;
 import jenkins.slaves.DeprecatedAgentProtocolMonitor;
 import org.eclipse.jetty.security.HashLoginService;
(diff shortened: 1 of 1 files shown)


oleg-nenashev (changes requested):
Hi @varyvol . I disagree that we do not want warnings in the build logs at all. IMO we should have warnings which are sensitive to users and which may impact the build result.  

JENKINS-54425 is about a particular message, about which users cannot do much. And this message is generally harmless

varyvol:
@oleg-nenashev I have downgraded the level only for [the logger logging such message](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/ClassicPluginStrategy.java#L867).

fcojfernandez (approved):

setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsfileRunnerRule.java
```diff
@@ -81,6 +82,7 @@ public void before() throws Throwable {
      */
     private void setLogLevels() {
         Logger.getLogger("").setLevel(Level.WARNING);
+        Logger.getLogger(ClassicPluginStrategy.class.getName()).setLevel(Level.SEVERE);
```
oleg-nenashev, line 86:
```suggestion
        // Prevent warnings for plugins with old plugin POM (JENKINS-54425)
        Logger.getLogger(ClassicPluginStrategy.class.getName()).setLevel(Level.SEVERE);
```

oleg-nenashev (approved):
There only two WARNING messages in the class. The second one (https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/ClassicPluginStrategy.java#L213) also does not make much sense

PR #50, opened 2018-12-20
Change tag name format

varyvol:
With the current configuration, the tags are generated such as parent-$version.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -124,7 +124,7 @@ THE SOFTWARE.
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-release-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


fcojfernandez (approved):

oleg-nenashev (commented):
Should work, but I would recommend appending config instead of overriding it. So it will help us to not lose settings from the parent POM . https://github.com/jenkinsci/pom/blob/master/pom.xml#L825-L830

jvz (approved):

varyvol:
@oleg-nenashev the effective-pom didn't display any override happening, but it doesn't hurt anyway changing the behavior.

oleg-nenashev (approved):

PR #51, opened 2018-12-20
[JENKINS-54353] Jenkinsfile Runner: Replace the JUnit implementation but something production-focused

fcojfernandez:
See [JENKINS-54353](https://issues.jenkins-ci.org/browse/JENKINS-54353)

Replacing JUnit code when the JFR is launched.
- Refactor to separate the JUnit classes (JenkinsfileRunnerRule) to a test module so it may be used in the future
- Split `JenkinsEmbedder` into two classes, one for the Jenkins engine and another one with the use of JUnit features

**Testing**
- [x] Manual
- [x] Demo
- [x] [CWP](https://github.com/jenkinsci/custom-war-packager/)
- [ ] [ci.jenkins.io-runner](https://github.com/oleg-nenashev/ci.jenkins.io-runner)


fcojfernandez:
Finally, I could not test using the ci.jenkins.io-runner. Even an execution using the demo in master is failing for me because of some dependencies error:
```
[ERROR] 
[ERROR] Try downloading the file manually from the project website.
[ERROR] 
[ERROR] Then, install it using the command: 
[ERROR]     mvn install:install-file -DgroupId=org.jboss.marshalling -DartifactId=jboss-marshalling-river -Dversion=1.4.12.jenkins-3 -Dpackaging=hpi -Dfile=/path/to/file
[ERROR] 
[ERROR] Alternatively, if you host your own repository you can deploy the file there: 
[ERROR]     mvn deploy:deploy-file -DgroupId=org.jboss.marshalling -DartifactId=jboss-marshalling-river -Dversion=1.4.12.jenkins-3 -Dpackaging=hpi -Dfile=/path/to/file -Durl=[url] -DrepositoryId=[id]
[ERROR] 
[ERROR] 
[ERROR]   org.jboss.marshalling:jboss-marshalling-river:hpi:1.4.12.jenkins-3
[ERROR] 
[ERROR] from the specified remote repositories:
[ERROR]   cloudbees-internal (https://nexus-internal.cloudbees.com/content/groups/staged, releases=true, snapshots=true)
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException

```

app/pom.xml
```diff
@@ -70,6 +70,11 @@
       <artifactId>payload</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
```
oleg-nenashev: The PR description says "can be used in the future", but actually it is bundled into the executable along with JUnit. I doubt we want it

test/src/test/java/io/jenkins/jenkinsfile/runner/JenkinsTestEmbedder.java
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package io.jenkins.jenkinsfile.runner;
```
oleg-nenashev, line 25: maybe it makes sense to introduce a new package here. E.g. `io.jenkins.jenkinsfile.runner.junit` or so

setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsEmbedder.java
```diff
@@ -136,11 +131,6 @@
 
     protected List<JenkinsRecipe.Runner> recipes = new ArrayList<JenkinsRecipe.Runner>();
 
-    /**
-     * Number of seconds until the test times out.
-     */
-    public int timeout = Integer.getInteger("jenkins.test.timeout", System.getProperty("maven.surefire.debug") == null ? 180 : 0);
```
oleg-nenashev, line 142: Was doing nothing anyway

setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsfileRunnerLauncher.java
```diff
@@ -0,0 +1,137 @@
...
+     * This was initially motivated by SLF4J leaving gnarly messages.
+     * The whole JVM is going to die anyway, so we don't really care about cleaning up anything nicely.
+     */
+    @Override
+    public void after() throws Exception {
+        jenkins = null;
+        super.after();
+    }
+
+    //TODO: add support of timeout
```
oleg-nenashev, line 107: Created https://issues.jenkins-ci.org/browse/JENKINS-55295

```suggestion
    //TODO: add support of timeout (JENKINS-55295)
```

oleg-nenashev (changes requested):
Thanks @fcojfernandez !

I doubt we really need to keep the JUnit implementation codebase in the `master` branch. It may make sense to put it to a branch. In any case, it should not be bundled into the JFR binaries for now.

setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsfileRunnerLauncher.java
```diff
@@ -0,0 +1,137 @@
...
+        localPort = -1;
+
+        setPluginManager(new PluginManagerImpl(context.getServletContext(),pluginsDir));
+
+        return context.getServletContext();
+    }
+
+    @Override
+    public void recipe() throws Exception {
+        // TODO
```
varyvol: Do you expect more changes here?

setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsfileRunnerLauncher.java
```diff
@@ -0,0 +1,137 @@
...
+
+    //TODO: add support of timeout
+    /**
+     * Launch the Jenkins instance
+     * No time out and no output message
+     */
+    public int launch(Bootstrap bootstrap) throws Throwable {
+        int returnCode = -1;
+        Thread t = Thread.currentThread();
+        String o = t.getName();
```
varyvol: Maybe a more descriptive name for the variable?

test/src/test/java/io/jenkins/jenkinsfile/runner/JenkinsTestEmbedder.java
```diff
@@ -0,0 +1,77 @@
...
+        env = new TestEnvironment(testDescription);
+        return this;
+    }
+
+    /**
+     * Called during the {@link #before()} to give a test case an opportunity to
+     * control the test environment in which Hudson is run.
+     */
+    @Override
+    public void recipe() throws Exception {
```
varyvol, line 57: Weren't recipes used previously to setup the Jenkins environment where to run the Pipeline?

setup/src/main/java/io/jenkins/jenkinsfile/runner/App.java
```diff
@@ -1,36 +1,16 @@
...
-                Class<?> c = cl.loadClass("io.jenkins.jenkinsfile.runner.Runner");
-                returnCode[0] = (int)c.getMethod("run",Bootstrap.class).invoke(c.newInstance(),bootstrap);
-            }
-        }, Description.createSuiteDescription("main"));
-
-        s.evaluate();
+        JenkinsfileRunnerLauncher launcher = new JenkinsfileRunnerLauncher(bootstrap.warDir, bootstrap.pluginsDir);
 
-        return returnCode[0];
+        return launcher.launch(bootstrap);
```
varyvol: What do you think of something like?

`
JenkinsfileRunnerLauncher launcher = new JenkinsfileRunnerLauncher(bootstrap);
launcher.launch();
`

app/pom.xml
fcojfernandez: you're right! I've added there for my earliest development and I forgot to remove it from here. It's actually not used.

setup/src/main/java/io/jenkins/jenkinsfile/runner/App.java
fcojfernandez: Sounds good to me!

setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsfileRunnerLauncher.java
fcojfernandez: I just forgot to change the comment. It's not needed for the launcher, but it could be useful if we add functional tests.

setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsEmbedder.java
fcojfernandez: Indeed. But I think it's a good idea to keep it in the Test class just in case in the future we could add this feature, so I've moved it. If we finally address JENKINS-55295, we just can move it back here again or look for another approach for production environment.

setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsfileRunnerLauncher.java
fcojfernandez: It was copied for an earlier class, but I agree.

test/src/test/java/io/jenkins/jenkinsfile/runner/JenkinsTestEmbedder.java
fcojfernandez: No class had the annotation `@JenkinsRecipe`, so they were not previously used at all. They can be used that way, but it's not actually used.

test/src/test/java/io/jenkins/jenkinsfile/runner/JenkinsTestEmbedder.java
fcojfernandez: I didn't changed the package since it's a test class in a test module and for me it's clear. However, if you both think it's better to add that new package, I will be happy to change it. Any though @varyvol ?

varyvol: I would leave as it is. It's unclear yet if this will remain or will be removed, and even if it remains it won't be used at the same time than the new Embedder.

varyvol (approved):

fcojfernandez:
@oleg-nenashev I agree with you and I have doubts about keeping the JUnit classes, but I feel more confortable if we keep them for now and move/remove them once we address the test-framework and the automated test for the JFR. Does it make sense for you?

oleg-nenashev:
There is a commit history for a reason. You can always recover the code
from it.

But I am not stronly against it as long as the code is not included into
the production

fcojfernandez:
It's just I'm not a fan of archaeology if it's not strictly needed, so to move or remove it when the epics for testing is addressed seems better moment to me. By the way, if you prefer me to do it now, I will be happy.
Regarding if this code is included into production, as you can see, the new module has only the main/test directory and I've already removed the dependency that I had kept by mistake.

vilacides (approved):
Given the two reviewers are out and all the comments have been addressed and the demos are passing, I am inclined to get this merged

PR #53, opened 2019-01-16
[JENKINS-54426] Fix failures in Java 11

varyvol:
[JENKINS-54426](https://issues.jenkins-ci.org/browse/JENKINS-54426)

Based in #41 

- Use platform classloader when JDK 9+ to be able to access platform classes.
- Include a demo that uses a base image with JDK11.

### DIFF
--- a/demo/cwp-jdk11/Makefile
+++ b/demo/cwp-jdk11/Makefile
@@ -14,6 +14,7 @@ clean:
 	rm -rf source
 	rm -rf out
 
+
 .copySource:
 	rsync -av --exclude='demo' $(shell pwd)/../../ $(shell pwd)/source
 
@@ -44,4 +45,6 @@ build: .copySource .build .buildImage
 buildInDocker: .copySource .buildInDocker .buildImage
 
 run:
-	docker run --rm -v $(shell pwd)/Jenkinsfile:/workspace/Jenkinsfile ${JENKINSFILE_RUNNER_TAG}
+	docker run --rm -v $(shell pwd)/Jenkinsfile:/workspace/Jenkinsfile \
+	-e JAVA_OPTS="-p /usr/share/jenkins/ref/java_cp/jaxb-api.jar:/usr/share/jenkins/ref/java_cp/javax.activation.jar --add-modules java.xml.bind,java.activation,java.sql -cp /usr/share/jenkins/ref/java_cp/jaxb-impl.jar:/usr/share/jenkins/ref/java_cp/jaxb-core.jar" \
+	${JENKINSFILE_RUNNER_TAG}
--- a/demo/cwp-jdk11/packager-config.yml
+++ b/demo/cwp-jdk11/packager-config.yml
@@ -42,7 +42,7 @@ plugins:
   - groupId: "org.jenkins-ci.plugins.workflow"
     artifactId: "workflow-support"
     source:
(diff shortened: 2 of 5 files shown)


demo/cwp-jdk11/Makefile
```diff
@@ -0,0 +1,50 @@
...
+	docker build -t ${JENKINSFILE_RUNNER_TAG} -f $(shell pwd)/out/tmp/output/Dockerfile $(shell pwd)/out/tmp/output/	
+
+build: .copySource .build .buildImage
+
+# Slow, but it does not need anything except Docker
+buildInDocker: .copySource .buildInDocker .buildImage
+
+run:
+	docker run --rm -v $(shell pwd)/Jenkinsfile:/workspace/Jenkinsfile \
+	-e JAVA_OPTS="-p /usr/share/jenkins/ref/java_cp/jaxb-api.jar:/usr/share/jenkins/ref/java_cp/javax.activation.jar --add-modules java.xml.bind,java.activation,java.sql -cp /usr/share/jenkins/ref/java_cp/jaxb-impl.jar:/usr/share/jenkins/ref/java_cp/jaxb-core.jar" \
```
varyvol, line 49: $JAVA_OPTS to be provided by the jenkinsfile-runner binary to the Java command.

demo/cwp-jdk11/packager-config.yml
```diff
@@ -0,0 +1,84 @@
...
+    source:
+      version: "2.30"
+  - groupId: "org.jenkins-ci.plugins.workflow"
+    artifactId: "workflow-step-api"
+    source:
+      version: "2.14"
+  - groupId: "org.jenkins-ci.plugins.workflow"
+    artifactId: "workflow-support"
+    source:
+      version: "3.0" # TO-DO: Maybe this should be into payload-dependencies ?
```
varyvol, line 45: It's not in the ["base demo"](https://github.com/jenkinsci/jenkinsfile-runner/blob/master/demo/cwp/packager-config.yml) so probably should be ported to payload-dependencies module in a future JDK11 branch.

jvz (approved):
I find it amusing that you use Makefiles instead of Ant (the slightly better make but with more angles in the syntax).

demo/cwp-jdk11/Makefile
oleg-nenashev: Ideally we need to include it into the image, but we need to update CWP to do that (trivial change)

oleg-nenashev (approved):
So in #41 I was patching a wrong classloader, and then did not experiment with the proper one. Thanks @varyvol ! 

I think we need to somehow automate bundling JAVA_OPTS into the image, but it can be done later as a part of https://issues.jenkins-ci.org/browse/JENKINS-54906

demo/cwp-jdk11/packager-config.yml
oleg-nenashev: I think we can update the payload in both branches. Better to have a new major release in place OOTB

demo/cwp-jdk11/Makefile
varyvol: https://issues.jenkins-ci.org/browse/JENKINS-54906 for that.

fcojfernandez (approved):
Looks great

oleg-nenashev:
Thanks @varyvol !

PR #54, opened 2019-01-17
[JENKINS-54907] Prepare Docker image for JDK 11

varyvol:
[JENKINS-54907](https://issues.jenkins-ci.org/browse/JENKINS-54907)

Specify the proper JAVA_OPTS in the Docker image so that JFR works with JDK 11.

Again based in #41

PR #57, opened 2019-01-17
[JENKIN-54907] Include changes in logic besides Dockerfile

varyvol:
#54 did not include the changes in #53.

### DIFF
--- a/demo/cwp-jdk11/README.md
+++ b/demo/cwp-jdk11/README.md
@@ -3,7 +3,7 @@ Jenkinsfile Runner demo
 
 This demo demonstrates building of Jenkinsfile Runner Docker images
 with [Custom WAR Packager](https://github.com/jenkinsci/custom-war-packager/)
-using JDK 11 based Docker image.
+using a JDK 11 based Docker image.
 This demo bundles some plugins so that it demonstrates a Pipeline execution.
 
 To build the Docker image, run `make clean build`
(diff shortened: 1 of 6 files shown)


demo/cwp-jdk11/README.md
```diff
@@ -0,0 +1,16 @@
+Jenkinsfile Runner demo
+===
+
+This demo demonstrates building of Jenkinsfile Runner Docker images
+with [Custom WAR Packager](https://github.com/jenkinsci/custom-war-packager/)
+using a JDK 11 based Docker image.
+This demo bundles some plugins so that it demonstrates a Pipeline execution.
+
+To build the Docker image, run `make clean build`
```
jvz, line 9: Or just `make` since you made `all` an alias for those two targets.

jvz (approved):
LGTM

demo/cwp-jdk11/README.md
fcojfernandez: @jvz is right. We could refer to both in the README

fcojfernandez (approved):

demo/cwp-jdk11/README.md
varyvol: I will take this into account for future changes. I don't want to change it now because I want master and jdk11 branches not to differ except on the strictly necessary stuff.

PR #58, opened 2019-01-18
[JENKINS-54391] Create a minimum framework for testing Jenkinsfile Runner

fcojfernandez:
See [JENKINS-54391](https://issues.jenkins-ci.org/browse/JENKINS-54391)

This PR pretends to be the first approach using the test framework:
- It can be executed via maven: ` mvn package -Denvironment=test`
- or executing the makefile: `make -C tests`

Some initial test that can be used as guide have been added as well.
Since the test framework relies on shUnit2, all the test classes in the `test` module (which has been renamed as `tests`) have been removed.

Downstream of [PR#1](https://github.com/jenkinsci/jenkinsfile-runner-test-framework/pull/1)


tests/pom.xml
```diff
@@ -0,0 +1,52 @@
...
+  <parent>
+    <groupId>io.jenkins.jenkinsfile-runner</groupId>
+    <artifactId>parent</artifactId>
+    <version>1.0-beta-5-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>tests</artifactId>
+
+  <dependencies>
+    <dependency>
```
varyvol: Is this dependency necessary?

tests/tests.sh
```diff
@@ -0,0 +1,61 @@
...
+
+downloaded_cwp_jar="to_update"
+
+. $test_framework_directory/utilities/utils.inc
+. $test_framework_directory/utilities/cwp/custom-war-packager.inc
+. $test_framework_directory/utilities/jfr/jenkinsfile-runner.inc
+
+oneTimeSetUp() {
+  downloaded_cwp_jar=$(download_cwp "$test_framework_directory" "$CWP_version")
+  downloaded_cwp_jar=$(remove_string "$downloaded_cwp_jar" "CWP downloaded as ")
```
varyvol: Maybe we could change the response from "download_cpw()" to avoid having to do this?

tests/tests.sh
```diff
@@ -0,0 +1,61 @@
...
+}
+
+setUp() {
+  echo "Initializing the test case. Ignoring so far"
+}
+
+test_with_tag() {
+  jfr_tag=$(generate_docker_image_from_jar "$test_framework_directory" "$downloaded_cwp_jar" "$version" "$current_directory/test_resources/test_with_tag/packager-config.yml" "$jenkinsfile_runner_tag" | grep 'Successfully tagged')
```
varyvol: Why the grep?

tests/pom.xml
fcojfernandez: No. I forgot to remove it when I remove the test classes

tests/tests.sh
fcojfernandez: I though about that, but having only `echo "$cwp_jar_file"` looks odd to me

tests/tests.sh
fcojfernandez: Just for performance. The execution of the `java -jar cwp.jar` displays tons of logs and a variable in shell-script has a limited (although high) number of characters to store. That way the variable only will contain the text we are interested in.

tests/tests.sh
varyvol: But is that displayed in the logs during the test execution?

tests/tests.sh
varyvol: I see. Maybe we could think in redirecting the output to some kind of file to be consulted later? Something that can be done in the future though.

fcojfernandez: That's something to be thought once the hooks are developed because any concrete test depends on the framework's (library/hooks/helpers) design. For the moment, those tests are just a start point.

tests/tests.sh
fcojfernandez: Changed

varyvol (approved):

oleg-nenashev (commented):
It would be great to get it running within the pull request builder before it gets merged

fcojfernandez:
@oleg-nenashev those tests are just some examples and guidance. Maybe it makes more sense to integrate it to the PR-builder once we are addressing the real tests for the JFR repo.

tests/tests.sh
```diff
@@ -0,0 +1,60 @@
...
+
+oneTimeSetUp() {
+  downloaded_cwp_jar=$(download_cwp "$test_framework_directory" "$CWP_version")
+}
+
+setUp() {
+  echo "Initializing the test case. Ignoring so far"
+}
+
+test_with_tag() {
```
oleg-nenashev: AFAICT this method body should be a method of the test framework in https://github.com/jenkinsci/jenkinsfile-runner-test-framework/pull/1 . Looks to be a standard smoke test method

tests/tests.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+
```
oleg-nenashev, line 3: Should fail on command error by default . `set -e`

tests/tests.sh
```diff
@@ -0,0 +1,60 @@
...
+current_directory=$(pwd)
+test_framework_directory="$current_directory/jenkinsfile-runner-test-framework"
+sh_unit_directory="$test_framework_directory/shunit2"
+
+version="256.0-test"
+CWP_version="1.3"
```
oleg-nenashev, line 9: Why do you use old version here? 1.5 is the current version

oleg-nenashev: Ideally it should be using a default version set in the JFR Test Framework, just to K.I.S.S

tests/test_resources/test_with_default_tag/packager-config.yml
```diff
@@ -0,0 +1,80 @@
+bundle:
```
oleg-nenashev: IIRC the plan was to support multiple tests (Jenkinsfiles) for a single configuration. The current structure does not seem to reflect it.

pom.xml
```diff
@@ -51,10 +51,10 @@ THE SOFTWARE.
   <modules>
     <module>bootstrap</module>
     <module>setup</module>
-    <module>test</module>
```
oleg-nenashev, line 54: nice to see it was finally deleted though it is not related to this PR

tests/tests.sh
```diff
@@ -0,0 +1,60 @@
...
+jenkinsfile_runner_tag="jenkins-experimental/jenkinsfile-runner-test-image"
+
+downloaded_cwp_jar="to_update"
+
+. $test_framework_directory/utilities/utils.inc
```
oleg-nenashev, line 14: I think the framework should be sourced by a single command, maybe you could add a wrapper init script in https://github.com/jenkinsci/jenkinsfile-runner-test-framework/pull/1

tests/pom.xml
```diff
@@ -0,0 +1,44 @@
...
+            <version>1.6.0</version>
+            <executions>
+              <execution>
+                <id>test</id>
+                <phase>test</phase>
+                <goals>
+                  <goal>exec</goal>
+                </goals>
+                <configuration>
+                  <executable>make</executable>
```
oleg-nenashev, line 35: NIT: make is not really needed for the automated run. You could invoke tests directly

oleg-nenashev (commented):
I would like to discuss it before it gets merged.

tests/test_resources/test_with_default_tag/packager-config.yml
varyvol: Actually the paths for both the configuration and the Jenkinsfile are provided when calling the test, so why do you think it's not possible?

tests/tests.sh
varyvol: You mean it's not a real test?

fcojfernandez: IIUC, @oleg-nenashev you mean those tests are good enough to be smoke tests for the framework. My initial intention was to have the structure and one test that could be used as exemple. I'll put these tests in the framework and just add a totally fake test here that could be removed once the real tests are improved.

tests/tests.sh
fcojfernandez: It was just an initial example, but I will change it.
Regarding the default version, I'll see how to add it, but let's do it in a follow-up

tests/test_resources/test_with_default_tag/packager-config.yml
fcojfernandez: @oleg-nenashev it's just a test yaml as example. The real tests are not already implemented

tests/pom.xml
fcojfernandez: Yes, but since JFR and CWP use make as tool I though it could be a good idea to keep it for coherence. We can just change it and migrate the makefile to shell scripts in a follow up if all of us agree it's a good idea.

oleg-nenashev (approved):

PR #60, opened 2019-01-21
[JENKINS-55703] - Add a  “—runWorkspace” parameter to define the workspace 

oleg-nenashev:
Also, the new parameter is enabled by default for vanilla Docker images.

### DIFF
--- a/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
@@ -5,6 +5,7 @@ import org.kohsuke.args4j.CmdLineException;
 import org.kohsuke.args4j.CmdLineParser;
 import org.kohsuke.args4j.Option;
 
+import javax.annotation.CheckForNull;
 import javax.annotation.PostConstruct;
 import java.io.File;
 import java.io.IOException;
@@ -47,6 +48,15 @@ public class Bootstrap {
      */
     @Option(name = "-f", aliases = { "--file" }, usage = "Path to Jenkinsfile (or directory containing a Jenkinsfile) to run, default to ./Jenkinsfile.")
     public File jenkinsfile;
+
+    /**
+     * Workspace for the Run
+     */
+    @CheckForNull
+    @Option(name = "--runWorkspace", usage = "Path to the workspace of the run to be used within the node{} context. " +
+            "It applies to both Jenkins master and agents (or side containers) if any. " +
+            "Requires Jenkins 2.119 or above")
+    public File runWorkspace;
 
 
     public static void main(String[] args) throws Throwable {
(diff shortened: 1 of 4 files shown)


fcojfernandez (approved):

oleg-nenashev:
So the tricky thing here is to get it integrated with Custom WAR Packager. It would require bumping  minimum JFR requirement there (or to pass JAVA_OPTS) there instead

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
```diff
@@ -5,6 +5,7 @@
 import org.kohsuke.args4j.CmdLineParser;
 import org.kohsuke.args4j.Option;
 
+import javax.annotation.CheckForNull;
```
varyvol, line 8: You added FIndbugs but the annotation is the one from javax?

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
```diff
@@ -109,6 +119,15 @@ private void postConstruct() throws IOException {
                 installPlugin(shortname, version);
             }
         }
+
+        if (this.runWorkspace != null){
+            if (System.getProperty("jenkins.model.Jenkins.workspacesDir") != null) {
+                //TODO(oleg_nenashev): It would have been better to keep it other way, but made it as is to retain compatibility
+                System.out.println("Ignoring the --runWorkspace argument, because an explicit System property is set");
```
varyvol: Maybe we can provide the name of the system property in the message?

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
```diff
@@ -109,6 +119,15 @@ private void postConstruct() throws IOException {
                 installPlugin(shortname, version);
             }
         }
+
+        if (this.runWorkspace != null){
+            if (System.getProperty("jenkins.model.Jenkins.workspacesDir") != null) {
```
jvz: You have this property used multiple times. Want to refactor into a constant?

oleg-nenashev: no. but i will

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
oleg-nenashev: Yes. It comes from `com.google.code.findbugs`

varyvol (approved):

fcojfernandez (approved):
Re-approving

PR #65, opened 2019-02-10
[JENKINS-56079] Print the classpath in case of a ClassNotFoundException

carlossg:
When building projects on top of jenkinsfile-runner if you pass the wrong directories it just thows ClassNotFoundException with no hints to what's wrong

eg. https://github.com/carlossg/jenkinsfile-runner-lambda/blob/master/src/main/java/org/csanchez/jenkins/lambda/Handler.java#L172

### DIFF
--- a/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
@@ -12,12 +12,12 @@ import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
 import java.net.URI;
 import java.net.URL;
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
@carlossg could you please create a JIRA ticket and explain how you got into this situation in it?

varyvol:
Yes, I'm confused because that class is from the Jenkinsfile Runner itself, so it's not something that should fail.

carlossg:
https://issues.jenkins-ci.org/browse/JENKINS-56079

varyvol (commented):
@carlossg thanks! I was not aware of that kind of usage of the tool.

varyvol (approved):

fcojfernandez (approved):

oleg-nenashev (approved):
error message could be formatted better, but LGTM

PR #66, opened 2019-02-11
Reduce size of docker image

carlossg:
Do not repack the war
From 930MB to 814MB
Add a Dockerfile-dev to build image with locally build maven artifacts


Dockerfile
```diff
@@ -22,11 +22,22 @@ COPY --from=jenkinsfilerunner-mvncache /mavenrepo /mavenrepo
...
+ENV JENKINS_UC https://updates.jenkins.io
+USER root
+RUN mkdir -p /app /usr/share/jenkins/ref/plugins
+COPY --from=jenkins /app/jenkins /app/jenkins
+COPY --from=jenkins /usr/local/bin/install-plugins.sh /usr/local/bin/install-plugins.sh
+COPY --from=jenkins /usr/local/bin/jenkins-support /usr/local/bin/jenkins-support
 COPY plugins.txt /usr/share/jenkins/ref/plugins.txt
-RUN /usr/local/bin/install-plugins.sh < /usr/share/jenkins/ref/plugins.txt
+# empty war file to workaround https://github.com/jenkinsci/docker/pull/786
+RUN echo UEsFBgAAAAAAAAAAAAAAAAAAAAAAAA== | base64 -d > /usr/share/jenkins/jenkins.war && /usr/local/bin/install-plugins.sh < /usr/share/jenkins/ref/plugins.txt && rm -f /usr/share/jenkins/jenkins.war
```
oleg-nenashev: Could we just integrate a release of that?

carlossg: that was just merged, can use in the next release

oleg-nenashev:
It should land today

Dockerfile
```diff
@@ -22,11 +22,22 @@ COPY --from=jenkinsfilerunner-mvncache /mavenrepo /mavenrepo
 ADD . /jenkinsfile-runner
 RUN cd /jenkinsfile-runner && mvn package
 
-FROM jenkins/jenkins:${JENKINS_VERSION}
+FROM jenkins/jenkins:${JENKINS_VERSION} as jenkins
 USER root
-RUN mkdir /app && unzip /usr/share/jenkins/jenkins.war -d /app/jenkins
+# Delete big files not needed
+RUN mkdir /app && unzip /usr/share/jenkins/jenkins.war -d /app/jenkins && \
+  rm -rf /app/jenkins/scripts /app/jenkins/jsbundles /app/jenkins/css /app/jenkins/images /app/jenkins/help /app/jenkins/WEB-INF/detached-plugins /app/jenkins/winstone.jar /app/jenkins/WEB-INF/jenkins-cli.jar /app/jenkins/WEB-INF/lib/jna-4.5.2.jar
```
varyvol, line 29: What happens if a plugin relies on a detached plugin?

carlossg: the install-plugins script would install it

varyvol (approved):

Dockerfile
oleg-nenashev: I am not sure about that: https://github.com/jenkinsci/docker/blob/master/install-plugins.sh#L148

carlossg: there's no war in the last dockerfile `FROM` step where install-plugins is run

oleg-nenashev (changes requested):
Needs to be integrated with 2.164 changes. https://github.com/jenkinsci/jenkinsfile-runner/pull/66/files#r255445975

oleg-nenashev (approved):

Dockerfile
```diff
@@ -22,9 +22,19 @@ COPY --from=jenkinsfilerunner-mvncache /mavenrepo /mavenrepo
...
+
+FROM openjdk:8-jdk
```
varyvol, line 31: How will this work when the specified JENKINS_VERSION (which is a parameter) is modified to something needing Java 11?

oleg-nenashev: Unfortunately Docker is not that good at multi-platform support as it. We would need to add a new Dockerfile or to somehow implement Dockerfile generation from scripts and then to move the release flow to ci.jenkins.io and Trusted CI

varyvol: Yes, but that situation is working right now because the proper JDK is being set by the Jenkins Docker image itself right?

oleg-nenashev: No, JDK comes from the base image which is openjdk:8-jdk

oleg-nenashev (approved):
It will work as it now, but the same approach cannot be easily applied to Custom WAR Packager. Dockerfile pattern there will need an update

PR #68, opened 2019-02-11
Fix demo/cwp-jdk11

fcojfernandez:
After executing the demo for cwp-jdk11, I've got the following error message:
```
Error occurred during initialization of boot layer
java.lang.module.FindException: Module java.xml.bind not found
Makefile:48: recipe for target 'run' failed
make: *** [run] Error 1
```
JAXB is not part of Jenkins anymore since it's going to be detached to a new plugin. This PR pretends to fix the demo

### DIFF
--- a/demo/cwp-jdk11/packager-config.yml
+++ b/demo/cwp-jdk11/packager-config.yml
@@ -14,14 +14,14 @@ buildSettings:
       source:
         dir: ./source
     docker:
-      base: "jenkins/jenkins:jdk11"
+      base: "jenkins/jenkins:2.164-jdk11"
       tag: "jenkins-experimental/jenkinsfile-runner-demo"
       build: false
 war:
   groupId: "org.jenkins-ci.main"
   artifactId: "jenkins-war"
   source:
-    version: "2.158"
+    version: "2.164"
 plugins:
   - groupId: "org.jenkins-ci.plugins.workflow"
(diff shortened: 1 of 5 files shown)


demo/cwp-jdk11/packager-config.yml
```diff
@@ -14,14 +14,14 @@ buildSettings:
       source:
         dir: ./source
     docker:
-      base: "jenkins/jenkins:jdk11"
+      base: "jenkins/jenkins:2.164-jdk11"
```
oleg-nenashev, line 17: All other demos should be probably updated to pin base image versions as well

oleg-nenashev (approved):

oleg-nenashev (changes requested):
TBH I do not think that demos need to be updated to the latest weekly. https://github.com/jenkinsci/jenkinsfile-runner/pull/68#discussion_r255603849 suggests to have fixed versions, that's it. But all other demos in this repo actually have fixed versions. If you update them, it is better to wait till Wednesday and update to 2.150.3

varyvol (approved):

fcojfernandez:
@oleg-nenashev I misunderstood your comment. I've updated to the current LTS and once the cwp-jdk11 demo can be updated to a LTS version we can align all the demos

oleg-nenashev (approved):

PR #26, opened 2018-05-31
Allow for passing worfklow job parameters

toshke:
## Feature description

Idea is that `jenkinsfile-runner` can be used for testing of shared libraries with different parameters passed to test pipelines, e.g. same pipeline using shared library functionality should fail or succeed, depending on parameters passed to pipeline job. Parameters are passed using `-a` or `--args`, inspired by @ndeloof comment on https://github.com/kohsuke/jenkinsfile-runner/issues/13

Also, another CLI switch (`-ns or  --no-sandbox`) is added in order to disable sandboxing. Default behaviour is not changed however. 

## Implementation notes

Parameters are passed in following format
`-params="param1=value1&params2=value2"`

CLI parsing is done in such way that parameters are separated by ampersand symbol `&`, hence disallowing this symbol to be used in parameter value. There is safety check  (e.g. passing `-params=value` will result in appropriate error message)

```
root@194280c42eb8:/src# /app/bin/jenkinsfile-runner -w /app/jenkins -p /usr/share/jenkins/ref/plugins -f test -a value
Exception in thread "main" java.lang.RuntimeException: parameters massed be passed in param1=value1&param2=value2 format
        at io.jenkins.jenkinsfile.runner.bootstrap.Bootstrap.lambda$postConstruct$0(Bootstrap.java:136)
        at java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:948)
        at java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:580)
        at io.jenkins.jenkinsfile.runner.bootstrap.Bootstrap.postConstruct(Bootstrap.java:135)
        at io.jenkins.jenkinsfile.runner.bootstrap.Bootstrap.main(Bootstrap.java:81)

```

PR #69, opened 2019-02-11
Check for null in bootstrap.workflowParameters

carlossg:
It is defined with `@CheckForNull`


oleg-nenashev:
Yes, it was one of the topics in #26 .

oleg-nenashev (approved):

varyvol:
Wasn't this solved by https://github.com/jenkinsci/jenkinsfile-runner/pull/26/files#diff-ee2e34830e8899b2776e794c3a90a194R145 ?

carlossg:
there's nothing that forces calling `postConstruct` so you can end with NPE anyway

varyvol (approved):

PR #70, opened 2019-02-12
[JENKINS-54352] Create basic functional tests for Jenkinsfile Runner

fcojfernandez:
See [JENKINS-54352](https://issues.jenkins-ci.org/browse/JENKINS-54352)

**Tests for "vanilla" images** (standard ones) which we do not have:
- "Hello world" smoke test using a scripted pipeline (probably similar to this).
- "Hello world" smoke test using a declarative pipeline.

**Tests for CWP-produced images** mimicking the demos


tests/tests-vanilla.sh
```diff
@@ -0,0 +1,41 @@
...
+
+  cd "$src_directory"
+  result=$(docker build -t "$jenkinsfile_runner_tag" --no-cache . | grep 'Successfully tagged')
+  execution_should_success "$?" "$jenkinsfile_runner_tag" "$result"
+
+  cd "$current_directory"
+}
+
+test_scripted_pipeline() {
+  result=$(docker run --rm -v "$current_directory/test_resources/vanilla-images/test_scripted_pipeline":/workspace "$jenkinsfile_runner_tag" -a "param1=Hello" -a "param2=value2")
```
varyvol, line 26: Why not using the "run_jfr_docker_image" function? Are you missing any functionality?

tests/tests-cwp.sh
```diff
@@ -0,0 +1,32 @@
...
+jenkinsfile_runner_tag="jenkins-experimental/jenkinsfile-runner-test-image"
+downloaded_cwp_jar="to_update"
+
+. $test_framework_directory/init-jfr-test-framework.inc
+
+oneTimeSetUp() {
+  rm -rf "$working_directory"
+  mkdir -p "$working_directory"
+  downloaded_cwp_jar=$(download_cwp "$working_directory")
+  jfr_tag=$(execute_cwp_jar_and_generate_docker_image "$working_directory" "$downloaded_cwp_jar" "$version" "$current_directory/test_resources/cwp-produced-images/config/packager-config.yml" "$jenkinsfile_runner_tag" | grep 'Successfully tagged')
```
varyvol, line 18: I think you don't need to save the result and provide it to "execution_should_success". If you don't specify anything, it will gather the logs automatically.

tests/tests-cwp.sh
```diff
@@ -0,0 +1,32 @@
...
+  execution_should_success "$?" "$jenkinsfile_runner_tag" "$jfr_tag"
+}
+
+test_cwp() {
+  run_jfr_docker_image "$jenkinsfile_runner_tag" "$current_directory/test_resources/cwp-produced-images/test_cwp/Jenkinsfile"
+  jenkinsfile_execution_should_succeed "$?"
```
varyvol, line 24: Maybe checking for the "Evergreen ..." string?

tests/tests-cwp-jdk11.sh
```diff
@@ -0,0 +1,27 @@
...
+  rm -rf "$working_directory"
+  mkdir -p "$working_directory"
+  downloaded_cwp_jar=$(download_cwp "$working_directory")
+  jfr_tag=$(execute_cwp_jar_and_generate_docker_image "$working_directory" "$downloaded_cwp_jar" "$version" "$current_directory/test_resources/cwp-produced-images/config/packager-config-jdk11.yml" "$jenkinsfile_runner_tag" | grep 'Successfully tagged')
+  execution_should_success "$?" "$jenkinsfile_runner_tag" "$jfr_tag"
+}
+
+test_cwp-jdk11() {
+  run_jfr_docker_image "$jenkinsfile_runner_tag" "$current_directory/test_resources/cwp-produced-images/test_cwp-jdk11/Jenkinsfile"
+  jenkinsfile_execution_should_succeed "$?"
```
varyvol, line 24: Same, maybe checking for the return string.

tests/tests-vanilla.sh
fcojfernandez: Because I want to add `-a "param1=Hello" -a "param2=value2"`. There is currently an opened PR in the test framework (https://github.com/jenkinsci/jenkinsfile-runner-test-framework/pull/15) to allow it.Doing it this way we can add checks to test https://github.com/jenkinsci/jenkinsfile-runner/pull/26, which is not tested

tests/tests-cwp.sh
fcojfernandez: I tried but it didn't work. I'll give it a try again

tests/tests-cwp.sh
fcojfernandez: Good idea!

tests/tests-vanilla.sh
varyvol: Doesn't the changes introduced [here](https://github.com/jenkinsci/jenkinsfile-runner-test-framework/pull/13/files#diff-93253b70ce0abb644c69126e61809ff5R16) help? Or does the order affect?

tests/tests-cwp.sh
fcojfernandez: Now I remember why I had to do it this way. The log only is created when the executed function is `run_jfr_docker_image`, that is not the case.

tests/tests-vanilla.sh
fcojfernandez: The order affects: "$3" is an option for the `docker run` command, not for the execution of the Jenkinsfile-runner

varyvol (approved):

oleg-nenashev (approved):
LGTM

PR #71, opened 2019-02-15
[JENKINS-54352] Test negative scenarios

fcojfernandez:
See [JENKINS-54352](https://issues.jenkins-ci.org/browse/JENKINS-54352)

Adds test for negative scenarios:

- Set of plugins where some versions collide
- Jenkinsfile with wrong syntax
- Failure in the execution
- Pipeline execution hangs

PR #81, opened 2019-02-25
Fix CI execution

fcojfernandez:
In #71 I added some tests to check negative scenarios. One of them is testing a wrong plugin configuration (`test_plugin_versions_collision`), so after the full execution I tried to delete the generated docker image in the `oneTimeTearDown` function. It cannot be done in the CI because if there are more processes on execution making use of this image, the `oneTimeTearDown` function fails. Let's remove it.

The error log we get:
```
Error response from daemon: conflict: unable to delete 48dd0ab99acc (cannot be forced) - image is being used by running container f9b448838c51

ASSERT:Unknown failure encountered running a test
```
The failing function:
```
oneTimeTearDown() {
  # remove docker with invalid configuration
  docker_id=$(docker images -q "$jenkinsfile_runner_invalid_tag")
  docker rmi -f "$docker_id"
}
```

### DIFF
--- a/tests/tests-negative-scenarios.sh
+++ b/tests/tests-negative-scenarios.sh
@@ -85,6 +85,11 @@ private_execution_after_timeout() {
 
 oneTimeTearDown() {
   # force docker termination in case it is still executing
+  docker_id=$(docker ps -aqf "name=$jenkinsfile_runner_valid_tag")
+  if [ ! -z "$docker_id" ]
+  then
+    docker stop -t 1 "$docker_id"
+  fi
   docker_id=$(docker ps -aqf "name=$jenkinsfile_runner_invalid_tag")
   if [ ! -z "$docker_id" ]
   then
@@ -92,6 +97,11 @@ oneTimeTearDown() {
   fi
 
   # remove docker with invalid configuration
+  docker_id=$(docker images -q "$jenkinsfile_runner_valid_tag")
+  if [ ! -z "$docker_id" ]
+  then
+    docker rmi -f "$docker_id"
+  fi
   docker_id=$(docker images -q "$jenkinsfile_runner_invalid_tag")
   if [ ! -z "$docker_id" ]
   then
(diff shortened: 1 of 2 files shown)


oleg-nenashev:
Could we add the container termination to the method? IIUC the code, otherwise we may have a runaway container during the execution

varyvol:
Why would other process be using this image? Because of parallel tests?

fcojfernandez:
@oleg-nenashev @varyvol the problem is when there is more than one executing jobs at the same time. All of them are using the same value for `jenkinsfile_runner_invalid_tag`, so the docker image cannot be removed without forcing it and that can make other executions fail. Maybe a timestamp?

varyvol:
@fcojfernandez it sounds weird to me because the call is made inside a 'node' statement, so in theory and as per my understanding, the node where both executions happen should be different.

fcojfernandez:
@varyvol I didn't realise that point. Now I understand better @oleg-nenashev 's comment. Have a look at my last commit, please

tests/tests-negative-scenarios.sh
```diff
@@ -53,7 +53,8 @@ test_pipeline_execution_fails() {
 
 test_pipeline_execution_is_unstable() {
   run_jfr_docker_image "$jenkinsfile_runner_valid_tag" "$current_directory/test_resources/negative-scenarios/test_pipeline_execution_is_unstable/Jenkinsfile"
-  jenkinsfile_execution_should_be_unstable "$?"
+  execution_success "$?"
```
varyvol, line 56: Isn't this checking the same?

fcojfernandez: `jenkinsfile_execution_should_be_unstable` checks one more content in the logs, but after checking some executions in PRs and master, I've seen it could cause some flakiness, so I feel more confortable if we prevent that.

varyvol: Fair enough!

varyvol (approved):

varyvol:
I have approved because the code looks good to me, but CI is still failing :(

fcojfernandez:
My last commit seems to work. My guess is that the docker image that cannot be removed was one of the common parents for both images used in the tests.

varyvol (approved):

oleg-nenashev (approved):
Whatever works. But the final implementation should really ne a part of the JFR Test Framework, right?

fcojfernandez:
@oleg-nenashev maybe a refactoring task can be a good idea

PR #87, opened 2019-02-25
[JENKINS-55227] Add a debug flag to Jenkinsfile Runner

fcojfernandez:
See [JENKINS-55227](https://issues.jenkins-ci.org/browse/JENKINS-55227)

This PR pretends to add a debug flag so it's possible to debug Jenkinsfile Runner using the docker image.

- Add an alternative script to execute Jenkinsfile Runner which reads the `DEBUG` environment variable to enable or not the remote debug
- Update the Dockerfile to use it
- Update the documentation


Dockerfile
```diff
@@ -28,6 +28,7 @@ RUN mkdir /app && unzip /usr/share/jenkins/jenkins.war -d /app/jenkins
 COPY plugins.txt /usr/share/jenkins/ref/plugins.txt
 RUN /usr/local/bin/install-plugins.sh < /usr/share/jenkins/ref/plugins.txt
 COPY --from=jenkinsfilerunner-build /jenkinsfile-runner/app/target/appassembler /app
+COPY jenkinsfile-runner /app/bin
```
oleg-nenashev: keep this file in `sec/docker`?

oleg-nenashev (commented):
It is not clear to me why it needed copying the appasembler script instead of creating a new smoll script which would call the generated script

fcojfernandez:
@oleg-nenashev It's clearer for me having a unique script, but I see your point. I'm happy to change it

jenkinsfile-runner-launcher
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+if [ -n "$DEBUG" ] ; then
+  export JAVA_OPTS="-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005 $JAVA_OPTS"
+fi
+
+. /app/bin/jenkinsfile-runner
```
oleg-nenashev: add a line break?

oleg-nenashev (approved):

jenkinsfile-runner-launcher
fcojfernandez: :+1:

jenkinsfile-runner-launcher
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+if [ -n "$DEBUG" ] ; then
+  export JAVA_OPTS="-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005 $JAVA_OPTS"
```
varyvol, line 4: Would it be possible to specify this in the same call than the command so that we don't change the value of the environment variable?

DOCKER.md
```diff
@@ -31,3 +31,12 @@ $ docker run --rm -it -v $PWD/test:/workspace --entrypoint bash jenkinsfile-runn
 root@dec4c0f12478:/src# cp -r /app/jenkins /tmp/jenkins
 root@dec4c0f12478:/src# /app/bin/jenkinsfile-runner -w /tmp/jenkins -p /usr/share/jenkins/ref/plugins -f /workspace
 ```
+
+## Debug
+In case you want to debug Jenkinsfile Runner the recommendation is to use the image built following the steps mentioned in the section above.
+
+Then, set the `DEBUG` environment variable and expose the port where to connect the remote debug
```
varyvol, line 38: I think it's worth specifying that the port to be exposed is 5005, given it cannot be changed.

oleg-nenashev: It can be changed on the user side. `docker run -p 1234:5005 -e DEBUG=1`

jenkinsfile-runner-launcher
fcojfernandez: The command is generated by appassembler, so the only way to pass this arguments is using the environment variable:

```
exec "$JAVACMD" $JAVA_OPTS  \
  -classpath "$CLASSPATH" \
  -Dapp.name="jenkinsfile-runner" \
  -Dapp.pid="$$" \
  -Dapp.repo="$REPO" \
  -Dapp.home="$BASEDIR" \
  -Dbasedir="$BASEDIR" \
  io.jenkins.jenkinsfile.runner.bootstrap.Bootstrap \
  "$@"
```

DOCKER.md
fcojfernandez: Oleg's right. The internal port is always 5005, but users can specify any other port for their remote debug.

varyvol: Yes, I know, but I think it's worth mentioning the one that can't be changed.

jenkinsfile-runner-launcher
varyvol: Yes, but would something like "JAVA_OPTS=xxx /app/bin/jenkinsfile-runner" work?

DOCKER.md
fcojfernandez: Would it work for you? 
```suggestion
Then, set the `DEBUG` environment variable and expose the port 5005 where to connect the remote debug
```
In my opinion this might be ambiguous about the ports to expose/configure.

jenkinsfile-runner-launcher
fcojfernandez: I don't think so. That's changing the JAVA_OPTS equally. Or might I be missing something on your proposal?

DOCKER.md
varyvol: I prefer it but if you don't feel sure enough or think it's confusing, I'm fine with it too.

jenkinsfile-runner-launcher
varyvol: I meant the JAVA_OPTS to be changed permanently, but the export is only inside the script so you can forget about my comment.

varyvol (approved):

oleg-nenashev:
Oops, missed https://github.com/jenkinsci/jenkinsfile-runner/pull/87#discussion_r260214971 . Should I create a follow-up PR?

varyvol:
@oleg-nenashev no worries. Given it's not clear if my suggestion would be good or not, let's leave as is.

PR #99, opened 2019-03-19
[JENKINS-55642] Create experimental "vanilla" Docker packages for the Vanilla version of JFR

fcojfernandez:
See [JENKINS-55642](https://issues.jenkins-ci.org/browse/JENKINS-55642)

Using the `infra.withDockerCredentials`, this PR add the experimental Docker build/publish flow to the Jenkinsfile. The arg `JENKINS_VERSION` in Dockerfile has been updated to `2.164.1` so the built image is compatable with java 8 and java 11.


Jenkinsfile
```diff
@@ -84,3 +83,30 @@ demos['databound'] = {
 }
 
 parallel(demos)
+
+node('docker') {
```
oleg-nenashev, line 87: I am not sure you want to run publishing for pull requests except this one

Jenkinsfile
```diff
@@ -55,7 +55,6 @@ for (int i = 0; i < platforms.size(); ++i) {
 /* Execute our platforms in parallel */
 parallel(branches)
 
-
 stage('Verify demos')
```
varyvol: Unrelated but given we now have automated tests, should we removed this?

Jenkinsfile
```diff
@@ -84,3 +83,30 @@ demos['databound'] = {
...
+    def image
+    def imageName = 'jenkins/jenkinsfile-runner-experimental'
+    def imageTag
+
+    stage('Checkout') {
+        timestamps {
+            deleteDir()
+            checkout scm
+
+            sh 'git rev-parse HEAD > GIT_COMMIT'
```
varyvol: If you store the result from "checkout scm", the git commit is something you can ask to that (see https://jenkins.io/doc/pipeline/steps/workflow-scm-step).

Jenkinsfile
varyvol: Agree.

Jenkinsfile
fcojfernandez: I have no strong opinion. To keep this does not hurt, but it will make the builder take longer

Jenkinsfile
fcojfernandez: thanks!

Jenkinsfile
fcojfernandez: See my next commit and tell me if it's a fine for you

Jenkinsfile
varyvol: Yes, that's my reason actually.

Jenkinsfile
```diff
@@ -84,3 +83,32 @@ demos['databound'] = {
...
 parallel(demos)
+
+def branchName = currentBuild.projectName
+if (!branchName.startsWith('PR-')) {
+    node('docker') {
+        def image
+        def imageName = 'jenkins/jenkinsfile-runner-experimental'
+        def imageTag
+
+        stage('Checkout') {
```
varyvol: Probably better something actually referring to the experimental Docker image.

varyvol (approved):

Jenkinsfile
```diff
@@ -84,3 +83,32 @@ demos['databound'] = {
 }
 
 parallel(demos)
+
+def branchName = currentBuild.projectName
+if (!branchName.startsWith('PR-')) {
+    node('docker') {
+        def image
+        def imageName = 'jenkins/jenkinsfile-runner-experimental'
```
oleg-nenashev: It will not work ad you may expect. Experimental images must go to another DockerHub organization. 
See the naming recommendations in https://github.com/jenkinsci/jep/tree/master/jep/217#implementation-on-cijenkinsio

Jenkinsfile
```diff
@@ -84,3 +83,32 @@ demos['databound'] = {
...
+        def imageTag
+
+        stage('Build container') {
+            timestamps {
+                deleteDir()
+                def scmVars = checkout scm
+
+                def shortCommit = scmVars.GIT_COMMIT
+                imageTag = "${env.BUILD_ID}-build${shortCommit}"
```
oleg-nenashev: It will lead to a huge number of tags in DockerHub.
Could we start from ...

```suggestion
                imageTag = branchName.equals("master") ? "latest" : branchName
```

It would produce images for all branches in the main repo, but it will not spam new tags for each commit

Jenkinsfile
```diff
@@ -84,3 +83,32 @@ demos['databound'] = {
 }
 
 parallel(demos)
+
+def branchName = currentBuild.projectName
+if (!branchName.startsWith('PR-')) {
```
oleg-nenashev: Maybe it makes sense tostart from a master branch and "PR-99" so that you can actually test deployment in this PR.

oleg-nenashev (changes requested):
The current implementation does not match JEP-217, and it will not work properly AFAICT.

oleg-nenashev:
I have created https://cloud.docker.com/u/jenkins4eval/repository/docker/jenkins4eval/jenkinsfile-runner . Once the flow is adjusted to the process, it should have automatic deployment

I will also make sure to update JEP-217 to offer common patterns to avoid deployments from PRs

Jenkinsfile
fcojfernandez: Seeing your last comment, do you mean  
`def imageName = '${env.DOCKERHUB_ORGANISATION}/jenkinsfile-runner'`
?

Jenkinsfile
fcojfernandez: :+1:

Jenkinsfile
oleg-nenashev: Yes

fcojfernandez: ok!. Thanks for the clarification. I didn't understand correctly that excerpt then.

Jenkinsfile
oleg-nenashev: @fcojfernandez I do not think it has been resolved so far.
Also, I think it still makes sense to build an image for each PR as a part of the flow. We just should not push it

fcojfernandez: Right! It is not committed yet

fcojfernandez: Done! I just wanted to commit everything in a single commmit

oleg-nenashev:
```
[2019-03-20T11:13:13.449Z] [ERROR] Failed to execute goal org.apache.maven.plugins:maven-install-plugin:2.5.2:install (default-install) on project parent: Failed to install artifact io.jenkins.jenkinsfile-runner:parent:pom:1.0-beta-8-SNAPSHOT: /home/jenkins/.m2/repository/io/jenkins/jenkinsfile-runner/parent/1.0-beta-8-SNAPSHOT/parent-1.0-beta-8-SNAPSHOT.pom (No such file or directory) -> [Help 1]
```

I believe we still have issues with parallel workspace usage for PR builds

Jenkinsfile
```diff
@@ -83,4 +82,34 @@ demos['databound'] = {
     }
 }
 
-parallel(demos)
+parallel(demos)*/
+
+node('docker') {
+    def image
+    def imageName = "${env.DOCKERHUB_ORGANISATION}/jenkinsfile-runner"
```
olblak: This can't work, DOCKERHUB_ORGANISATION is only available inside "[withDockerCredentials](https://github.com/jenkins-infra/pipeline-library/blob/66a14842e350bd5364a1a70ca4287535cacc0390/vars/infra.groovy#L13)" and the job on ci.j,io fails because it tries to push "null/jenkinsfile-runner:PR-99"

Jenkinsfile
```diff
@@ -55,8 +54,8 @@ for (int i = 0; i < platforms.size(); ++i) {
 /* Execute our platforms in parallel */
 parallel(branches)
 
-
-stage('Verify demos')
+// TODO Restore once JENKINS-56632 is done
```
varyvol, line 57: I stated it in another PR but given it's currently commented, I would remove this. We have now the automated tests to do the verification job for us.

Jenkinsfile
```diff
@@ -83,4 +82,35 @@ demos['databound'] = {
...
+
+                def shortCommit = scmVars.GIT_COMMIT
+                imageTag = branchName.equals("master") ? "latest" : branchName
+                echo "Creating the container ${imageName}:${imageTag}"
+                image = docker.build("${imageName}:${imageTag}", '--no-cache --rm .')
+            }
+        }
+
+        // TODO: Update when PR-99 is merged
+        if (branchName.startsWith('master') || branchName.startsWith('PR-99')) {
```
varyvol, line 107: One doubt about this (sorry for not bringing it up before): what will happen with tags? Because I guess we want it published?

Jenkinsfile
fcojfernandez: I'm removing it in https://issues.jenkins-ci.org/browse/JENKINS-56632, which is already in progress, so I'll leave it here until then

Jenkinsfile
fcojfernandez: Since it's for experimental images, we only care about master.

varyvol (approved):
Fair enough!

Jenkinsfile
```diff
@@ -55,8 +54,8 @@ for (int i = 0; i < platforms.size(); ++i) {
 /* Execute our platforms in parallel */
 parallel(branches)
 
-
-stage('Verify demos')
+// TODO Restore once JENKINS-56632 is done
+/*stage('Verify demos')
```
oleg-nenashev, line 58: Needs to be reverted

fcojfernandez: @oleg-nenashev since it causes sometimes some workspace collisions, I decided commented. #101 is reverting the comment, although is in progress for the moment.

oleg-nenashev (approved):

PR #106, opened 2019-04-10
Introduced --runHome to specify the runs JENKINS_HOME

alxsap:
We are going to use the Jenkinsfile Runner in Docker to run ephemeral pipelines. In order to copy the status of the pipeline (final result, steps graph, logs,...), before the container vanishes, it would be easier to have a defined location (inside the container) for the used Jenkins home. And it is required to keep this folder after Jenkinsfile Runner finished (the folder will be gone anyway once the Docker container finished).

Currently the temp folder is created randomly and disposed after the run (e.g. `/tmp/jenkinsfileRunner.tmp/jfr67075404322545838.run`).

This pull request introduces a `--runHome` parameter.
> --runHome
> Path to the empty Jenkins Home directory to use for this run. If not specified a temporary directory will be created. Note that the folder specified via --runHome will not be disposed after the run.

## Side note

Is it the right approach to use/copy the files from `JENKINS_HOME/jobs/job/builds/1` or is there a better way to get all this information once the Jenkinsfile Runner finished?

Thanks and best regards,
Alex

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,7 +162,12 @@ Advanced options:
 * In the Vanilla `Dockerfile` the master workspace is mapped to `/build`.
   This directory can be exposed as a volume.
   The docker image generated with Custom War Packager maps the workspace to `/build` by default and it can be exposed as well.
-  However it is possible to override that directory if both the `-v` docker option and the `--runworkspace` Jenkinsfile Runner option are specified.
+  However it is possible to override that directory if both the `-v` docker option and the `--runWorkspace` Jenkinsfile Runner option are specified.
+* By default the JENKINS_HOME folder is randomly created and disposed afterwards. With the `--runHome` parameter in combination with the `-v` docker option it is possible to specify a folder.   
+  e.g. `docker run -v /local/Jenkinsfile:/workspace/Jenkinsfile -v /local/jenkinsHome:/jenkinsHome ${JENKINSFILE_RUNNER_IMAGE} --runHome /jenkinsHome`
+
+  This way you can access the build metadata in `<jenkinsHome>/jobs/job/builds/1`, like the build.xml, logs, and workflow data, even after the container finished.
+
 * The `-ns` and `-a` options can be specified and passed to the image in the same way as the command line execution.
 
 ## Docker build
--- a/README.md
+++ b/README.md
@@ -105,6 +105,10 @@ The executable of Jenkinsfile Runner allows its invocation with these cli option
 
 ```
  # Usage: jenkinsfile-runner -w [warPath] -p [pluginsDirPath] -f [jenkinsfilePath] [other options]
+ --runHome FILE          : Path to the empty Jenkins Home directory to use for
+                           this run. If not specified a temporary directory
+                           will be created. Note that the folder specified via
+                           --runHome will not be disposed after the run.
  --runWorkspace FILE     : Path to the workspace of the run to be used within
                            the node{} context. It applies to both Jenkins
                            master and agents (or side containers) if any.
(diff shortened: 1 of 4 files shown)


varyvol:
@alxsap the "runWorkspace" parameter is already available for that. See [this](https://github.com/jenkinsci/jenkinsfile-runner/blob/master/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java#L65).

[This Custom WAR packager test ](https://github.com/jenkinsci/custom-war-packager/blob/master/jenkinsfile-runner-tests/tests.sh#L114) can be used as an example.

alxsap:
Hello @varyvol, thanks for the feedback.

`--runWorkspace` is something different. This parameter sets the 'Path to the workspace of the run to be used within the node{} context', meaning the pipelines build workspace.

The new parameter `--runHome` sets the JENKINS_HOME to be used for the run, where besides other things the `jobs` folder is located. In a static Jenkins all jobs are located there (config, builds, logs, etc.).
In a one shot pipeline run, using jenkinsfile runner, there will be only one job - called `job` - with one build in the `builds` folder with number `1`. There (`JENKINS_HOME/jobs/job/builds/1`) we find the whole state of the pipeline run plus the logs.

Without the `--runHome` parameter those files are written to a random (unpredictable) temporary folder and this folder is disposed when jenkinsfile runner finishes.

I still believe this new parameter is useful and even required to get the mentioned information in a proper way.

Thanks and best regards,
Alex

varyvol:
@alxsap I understand your point now, thanks for the detailed explanation. A question comes to my mind: what is it that you are trying to retain for the build? Are you not able to save that into the workspace through any step in the Jenkinsfile?

I am asking because Jenkinsfile Runner already has several options which are not easy to understand. If there is a way of doing things without introducing extra ones, I think it benefits to everyone.

alxsap:
@varyvol We are going to provide something like "Pipeline as a Service" to other development teams. This service shall be Pipeline agnostic. Therefore including steps into pipelines to extract required data, push logs to a log service, etc. is not desirable. Instead we should get this information "from outside".

The files in the `/jobs/job/builds/1` folder currently look the most promising to us. Or can you recommend something better? We need the pipeline status, pipeline structure and status of steps, logs (of each step) - everything live while it is running and after it finished.

Thanks, Alex

varyvol:
It does sound like a normal Jenkins instance?

alxsap:
This "Pipeline as a Service" shall scale in a cloud native way. That's why it must not be a static Jenkins (see Jenkins Cloud Native SIG) but for each Pipeline execution we are going to run a Jenkinsfile runner in Kubernetes. And since the results are gone once the container is gone we need to collect the required results before the container ends.

oleg-nenashev (approved):
It looks good to me
@alxsap Could you please also add documentation patches into this pull request?

fcojfernandez (approved):
LGTM. And I agree with Oleg. Documentation should be updated before merging this PR.

thanks @alxsap !

oleg-nenashev (changes requested):
Just to follow-up on the previous comment, needs docs

README.md
```diff
@@ -105,6 +105,10 @@ The executable of Jenkinsfile Runner allows its invocation with these cli option
 
 ```
  # Usage: jenkinsfile-runner -w [warPath] -p [pluginsDirPath] -f [jenkinsfilePath] [other options]
+ --runHome FILE          : Path to the empty Jenkins Home directory to use for
```
oleg-nenashev, line 108: Does it have to be empty?

alexanderlink: That was my assumption since originally always a new temporary folder has been created. My [current implementation](https://github.com/jenkinsci/jenkinsfile-runner/pull/106/files#diff-d88d995828fb47f31944e53ec493d4bcR40) checks if the folder is empty.
I'm not sure what might happen if the folder is not empty and already contains (potentially conflicting) files.

oleg-nenashev (approved):
Let's keep it as is. If somebody wants this feature on a dirty workspace, it would be a separate story

fcojfernandez (approved):

PR #127, opened 2019-05-20
Changelog: Noting 1.0-beta-8

oleg-nenashev:
Just manually copied the Release Drafter output for now

CC @fcojfernandez


fcojfernandez (approved):
Is it worth to add #125 as internal change?

oleg-nenashev:
@fcojfernandez it is filtered out deliberately, because Jetty update was reverted. I could update the demo part tho

fcojfernandez (approved):
fine!

PR #143, opened 2019-07-30
Use Jenkins BOM to specify plugin versions + Update to 2.176.2

oleg-nenashev:
It should help a lot to manage the Pipeline dependencies in the project. Dependabot definitely struggles with that, so it should help a lot.

FTR https://github.com/jenkinsci/bom

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -60,7 +60,7 @@ THE SOFTWARE.
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <java.level>8</java.level>
-    <jenkins.version>2.164.3</jenkins.version>
+    <jenkins.version>2.176.2</jenkins.version>
     <jetty.version>9.4.5.v20170502</jetty.version>
     <!--TODO: Reenable once all the issues are fixed (JENKINS-57353)-->
     <findbugs.failOnError>false</findbugs.failOnError>
@@ -68,6 +68,13 @@ THE SOFTWARE.
 
   <dependencyManagement>
     <dependencies>
+      <dependency>
+        <groupId>io.jenkins.tools.bom</groupId>
+        <artifactId>bom</artifactId>
+        <version>${jenkins.version}</version>
+        <scope>import</scope>
+        <type>pom</type>
+      </dependency>
       <dependency>
         <groupId>commons-collections</groupId>
         <artifactId>commons-collections</artifactId>
@@ -83,12 +90,6 @@ THE SOFTWARE.
         <artifactId>slf4j-jdk14</artifactId>
         <version>1.7.25</version>
       </dependency>
-      <!-- Annotation engine which we want to keep at recent versions instead of ones bundled in the core -->
-      <dependency>
-        <groupId>org.jenkins-ci</groupId>
-        <artifactId>symbol-annotation</artifactId>
-        <version>1.17</version>
-      </dependency>
       <dependency>
         <groupId>org.jenkins-ci</groupId>
         <artifactId>annotation-indexer</artifactId>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -60,14 +60,21 @@ THE SOFTWARE.
...
     <!--TODO: Reenable once all the issues are fixed (JENKINS-57353)-->
     <findbugs.failOnError>false</findbugs.failOnError>
   </properties>
 
   <dependencyManagement>
     <dependencies>
+      <dependency>
+        <groupId>io.jenkins.tools.bom</groupId>
+        <artifactId>bom</artifactId>
+        <version>${jenkins.version}</version>
```
jglick:
```suggestion
        <version>2.176.2</version>
```

Read the manual.

pom.xml
```diff
@@ -99,6 +100,25 @@ THE SOFTWARE.
...
+      <dependency>
+        <groupId>org.jenkins-ci.plugins.workflow</groupId>
+        <artifactId>workflow-multibranch</artifactId>
+        <version>2.21</version>
+      </dependency>
+      <!--TODO(oleg-nenashev): Is it a valid use-case? Move to BOM if so -->
+      <dependency>
+        <groupId>org.jenkins-ci.plugins.workflow</groupId>
+        <artifactId>workflow-scm-step</artifactId>
+        <classifier>tests</classifier>
```
jglick, line 120: Why is this even here? I would delete it.

oleg-nenashev: I might. Need to check WTF it is first. I still discover various hacks related to the former usage of JTH inside JFR

pom.xml
oleg-nenashev: ack. Will be a permanent source of confusion for users.

jglick: Yeah...

oleg-nenashev:
```
[2019-07-30T16:30:21.084Z] INFO: Set new version for jenkinsfile-runner: 256.0-2019-07-30-SNAPSHOT

[2019-07-30T16:30:42.996Z] Exception in thread "main" java.io.IOException: Command failed with exit code 1: mvn clean install -DskipTests -Dfindbugs.skip=true -Denforcer.skip=true -Djenkins.version=2.150.2

[2019-07-30T16:30:42.996Z] 	at io.jenkins.tools.warpackager.lib.util.SystemCommandHelper.processFor(SystemCommandHelper.java:39)

[2019-07-30T16:30:42.996Z] 	at io.jenkins.tools.warpackager.lib.util.MavenHelper.run(MavenHelper.java:66)

[2019-07-30T16:30:42.996Z] 	at io.jenkins.tools.warpackager.lib.util.MavenHelper.run(MavenHelper.java:46)

[2019-07-30T16:30:42.996Z] 	at io.jenkins.tools.warpackager.lib.impl.Builder.buildIfNeeded(Builder.java:342)

[2019-07-30T16:30:42.996Z] 	at io.jenkins.tools.warpackager.lib.impl.Builder.build(Builder.java:193)

[2019-07-30T16:30:42.996Z] 	at io.jenkins.tools.warpackager.cli.Main.main(Main.java:56)

[2019-07-30T16:30:42.996Z] unable to prepare context: unable to evaluate symlinks in Dockerfile path: lstat /home/jenkins/platform_linux_PR-143_1/tests/.jenkinsfile-runner-test-framework/work-cwp/out/tmp/output/Dockerfile: no such file or directory

[2019-07-30T16:30:42.996Z] ASSERT:Unknown failure encountered running a test

[2019-07-30T16:30:42.996Z] 

```

uh oh

pom.xml
fcojfernandez: Which ones? I thought I had decoupled completely from JTH

oleg-nenashev: Yes an no. There is no dependencies on JTH specifically, but there are bits like https://issues.jenkins-ci.org/browse/JENKINS-55764 which likely require a follow-up.

pom.xml
```diff
@@ -60,14 +60,21 @@ THE SOFTWARE.
...
     <!--TODO: Reenable once all the issues are fixed (JENKINS-57353)-->
     <findbugs.failOnError>false</findbugs.failOnError>
   </properties>
 
   <dependencyManagement>
     <dependencies>
+      <dependency>
+        <groupId>io.jenkins.tools.bom</groupId>
+        <artifactId>bom</artifactId>
+        <version>2.176.2</version>
```
varyvol, line 75: So this takes the version in the specified BOM instead of the ones in this project POM right? Which version are included in such BOM, the ones necessary for Jenkins to run properly?

What about the dependencies that are not needed for Jenkins but for JFR itself?

oleg-nenashev: Basically the BOM provides a list of Pipeline plugins which are known to be compatible with each other. This is the main benefit, because maintaining Pipeline plugin dependencies is a nightmare now

Jenkins Core BOM is a separate story ATM, can link some internal chats in PM

oleg-nenashev:
```suggestion
        <!--WARNING: it is not a Jenkins core version. https://issues.jenkins-ci.org/browse/JENKINS-58770 -->
        <version>2.176.2</version>
```

varyvol:
I'm aware of the comfort it introduced regarding the pipeline plugins, but I'm not convinced because it will probably make things harder when we need to get specific versions (for example including bugfixes).

oleg-nenashev:
We can always override it if we need to do so at some point

varyvol:
@oleg-nenashev by including it in this project's POM? or how does it work?

oleg-nenashev:
Basically you can just set '<version>' explicitly to override dependency
management. So it is a one-line fix in 'payload-dependencies' if we want a
custom version

varyvol (approved):

jglick (approved):

PR #153, opened 2019-08-06
Move changelog to GitHub releases

oleg-nenashev:


oleg-nenashev:
There is a CI failure, but CI does not verifz documentation in any means

PR #167, opened 2019-09-18
Do not fail negative tests on image deletion failures

oleg-nenashev:
Just a tweak which should stabilize the CI for now until the underlying issue is fixed.
Alternate way is to just disable negative tests for now


oleg-nenashev:
CI was fixed by stabilizing Vanilla images and increasing timeouts. YAGNI

Issue #180, opened 2019-10-07
Add a CLI flag to print Jenkinsfile Runner version


oleg-nenashev:
Currently there is no CLI argument which would print a Jenkinsfile Runner version. It would be great to have one, e.g. `-v` and `--version`

mvarendorff:
Would love to do this! Could you give me a suggestion where to look for the flags?

The only place I found where CLI flags seem to be configured was the Bootstrap.java file which already contains an entry for -v/--version [here](https://github.com/jenkinsci/jenkinsfile-runner/blob/master/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java#L44). Is that the correct place? In that case would you suggest using a different name for the flag (and which)?

oleg-nenashev:
Uh oh, so we have an argument collision :( My preference would be to
introduce a breaking change while we are in Beta. The existing param could
be renamed to "-jenkinsVersion" or so.

Apoorv-Gaurav-Agarwal:
is this still available?

oleg-nenashev:
No, sorry. #182 has implemented it. I will make sure to create more issues today

oleg-nenashev:
It was released in https://github.com/jenkinsci/jenkinsfile-runner/releases/tag/1.0-beta-10 , thanks!

PR #182, opened 2019-10-07
#180 - Modify -v/--version flag to display the version + Use `-jv/--jenkinsVersion` for the former behavior

mvarendorff:
This PR changes the functionality of the previous -v/--version flag (which is now called -jv/--jenkins-version) to display the version of the jenkinsfile-runner.

A few notes:

 - I hope I got all correct occurences of the "old" -v flag in the readme
 - I did *not* change version yet as this is a breaking change and did not want to change the wrong things
 - I added a new variable for the version which would be hardcoded and would have to be changed. I couldn't find any config file and parsing the pom.xml surely is not the way to go :P If there is a proper place where I would get the version number from otherwise, I will gladly change the code to do so.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,28 +107,28 @@ The executable of Jenkinsfile Runner allows its invocation with these cli option
 
 ```
  # Usage: jenkinsfile-runner -w [warPath] -p [pluginsDirPath] -f [jenkinsfilePath] [other options]
- --runHome FILE          : Path to the empty Jenkins Home directory to use for
-                           this run. If not specified a temporary directory
-                           will be created. Note that the folder specified via
-                           --runHome will not be disposed after the run.
- --runWorkspace FILE     : Path to the workspace of the run to be used within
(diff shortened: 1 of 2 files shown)


mvarendorff:
I hope that the change I made is what you meant; that was the simplest I could come up with without starting to mess with maven plugins etc.

oleg-nenashev (approved):
It should do the job, thanks! I will test it manually and then merge it. CI is in a bad shape now unfortunately because of the Vanilla packaging problem

oleg-nenashev:
https://github.com/jenkinsci/jenkinsfile-runner/issues/185 ftr

oleg-nenashev (approved):
Everything should be aligned now, sorry for some non-related commits after the merge conflict resolution

mvarendorff:
No worries! Thanks for the opportunity to work on this and a PR that feels good for hacktoberfest :)

Issue #179, opened 2019-10-07
Add `--help` CLI argument which prints help


oleg-nenashev:
Currently the only way to get help is to pass invalid command. It would be great to support `-h` and `--help` in Jenkinsfile Runner CLI.

ImShakthi:
@oleg-nenashev I would like to work on this enhancement

oleg-nenashev:
@ImShakthi 👍 Just submit a pull request, thanks!

ImShakthi:
@oleg-nenashev Have created a PR #183 for this issue. Kindly let me know your feebacks.

oleg-nenashev:
Should be released soon

ImShakthi:
@oleg-nenashev Thank you :)

oleg-nenashev:
It was released in https://github.com/jenkinsci/jenkinsfile-runner/releases/tag/1.0-beta-10 . Thanks you for your contribution!

PR #183, opened 2019-10-08
#179 - Add '--help' CLI argument which prints help

ImShakthi:
Have added support to print help information to `jenkinsfile-runner` with `-h` or `--help`.


oleg-nenashev (approved):
Looks right, thanks!
I will integrate it once https://github.com/jenkinsci/jenkinsfile-runner/issues/185 is fixed

Issue #185, opened 2019-10-08
Vanilla Docker packaging downloads latest plugins instead of ones defined by payload-dependencies


oleg-nenashev:
Currently Docker packaging just explodes, because it pulls latest plugin dependencies instead of ones defined in https://github.com/jenkinsci/jenkinsfile-runner/blob/master/payload-dependencies/pom.xml .

I might need to integrate https://github.com/jenkinsci/custom-war-packager or https://github.com/jenkinsci/plugin-installation-manager-tool into the build process to retrieve a proper set of dependencies

PR #186, opened 2019-10-08
#185 Disable Vanilla image tests for now, image is broken

oleg-nenashev:
https://github.com/jenkinsci/jenkinsfile-runner/issues/185


oleg-nenashev:
#187 was a better solution (apart from the fact the Ci is still red)

Issue #175, opened 2019-10-06
Add Declarative Pipeline to the standard bundle


oleg-nenashev:
Currently Declarative Pipeline is not included in https://github.com/jenkinsci/jenkinsfile-runner/tree/master/payload-dependencies , but it is bundled into the Dockerfile's `plugins.txt` ti means there is no verification of the plugin until #172 is implemented with some test automation, and the plugin may just fail to start.

I suggest Declarative Pipeline is added to Jenkinsfile Runner explicitly, and that the `plugins.txt` hack is removed from the Vanilla Docker image

oleg-nenashev:
Fixed by #187

PR #187, opened 2019-10-08
#185, #175 - Manage Vanilla package jenkins.war and plugins through a Maven module to prevent startup failures

### LINKED ISSUE #185: Vanilla Docker packaging downloads latest plugins instead of ones defined by payload-dependencies

oleg-nenashev:
Fixes #185 . Basically, Vanilla package was at constant risk of binary conflicts due to the obsolete core and plugins. This change makes everything managed by Maven, and hence we get Enforcer checks and proper Dependabot integration. 

Docker images are also more simple now, but they can be still extended like before.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,6 @@ tests/.settings/
 tests/*.log
 demo/cwp-jdk11/out/
 demo/cwp-jdk11/source/
+vanilla-package/.classpath
+vanilla-package/.factorypath
+vanilla-package/target/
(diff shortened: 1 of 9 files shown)


oleg-nenashev:
I will go ahead and merge it, because the master branch is broken otherwise

PR #195, opened 2019-10-12
Disable Docker image deployment from ci.jenkins.io due to the unstable CI

oleg-nenashev:
DockerHub will be used instead for now


oleg-nenashev:
Merging without waiting for CI, syntax check has already passed

PR #198, opened 2019-10-13
Include the Plugin Installation Manager Tool into the Vanilla base image

### LINKED ISSUE #189: Add a "Customizing Jenkinsfile Runner" section to documentation
Since `jenkins4eval/jenkinsfile-runner` becomes closer to the state where it can be easily consumed, we could document extending the image

- [ ] Custom WAR packager flow is documented and referenced. https://jenkins.io/blog/2018/10/16/custom-war-packager/#jenkinsfile-runner-packaging
- [ ] A stub for a more classic image extension is added. It will be extended later

### LINKED ISSUE #190: Include Plugin Management Scripts into the Vanilla image
Eventiually I would like to make `jenkins4eval/jenkinsfile-runner` extensible in classic Docker ways, and plugin management is one of the common features there. The image used to include the plugin management scripts to include Declarative Pipeline, but it was removed in https://github.com/jenkinsci/jenkinsfile-runner/pull/187 .

It would be nice to restore bundling of plugin management scripts into the image. Then it can be documented for #189

mostafahussein:
This PR adds [jenkins-plugin-manager](https://github.com/jenkinsci/plugin-installation-manager-tool) to the final jenkinsfile-runner image which gives the ability to install custom plugins afterwards by using the image as base like the following:
```
FROM jenkins-runner-vanilla:latest
RUN cd /app/jenkins && jar -cvf jenkins.war *
RUN java -jar /app/bin/jenkins-plugin-manager.jar --war /app/jenkins/jenkins.war --plugin-file /usr/share/jenkins/ref/plugins.txt && rm /app/jenkins/jenkins.war
```
This will close #190 and close #189 after adding the documentation to this PR
Notes:
I couldn't make `jenkins-plugin-manager` to use the `/app/jenkins` directly like the runner command line can.
So can we make this possible in order to get rid of the extra line where I package the directory to war file?

Also, I have noticed this message:
```
Will use war file: /app/jenkins/jenkins.war
Unable to get version from war file
```
CC @stopalopa


oleg-nenashev (commented):
Thanks for your pull request!
Would it be possible to also add a small demo describing how to use it in downstream images?

Dockerfile
```diff
@@ -25,8 +25,10 @@ RUN mkdir /app && unzip /jenkinsfile-runner/vanilla-package/target/war/jenkins.w
 
 FROM openjdk:8-jdk
 ENV JENKINS_UC https://updates.jenkins.io
+ENV JENKINS_PM_VERSION 0.1-alpha-10
+ENV JENKINS_PM_URL https://github.com/jenkinsci/plugin-installation-manager-tool/releases/download/plugin-management-parent-pom-$JENKINS_PM_VERSION/jenkins-plugin-manager-$JENKINS_PM_VERSION.jar
```
oleg-nenashev, line 30: I wonder why it is located in POM instead of CLI. CC @stopalopa

mostafahussein: Do you mean that we should move these lines to be installed through the POM ?

oleg-nenashev: Sorry, missed the ping. No, no action required here. It is rather about the GitHub releases deployment

mostafahussein:
I will work on this PR today to resolve the conflict and add an example to the docs

mostafahussein:
@oleg-nenashev Dockerfile and docs updated. Can you please review it again ? note that the current conflict because I have modified steps that should  be added - if you approved - instead of the master version

oleg-nenashev (approved):
Documentation needs to be moved/enhanced before the release, but I will do it in a follow-up

Issue #196, opened 2019-10-13
Add Pipeline Utility Steps to the Vanilla Package


oleg-nenashev:
https://plugins.jenkins.io/pipeline-utility-steps is widely used for operations like YAML processing, and it would be great to have it as a part of the default package so that users do not need to extend the image just to get this plugin installed. 

Acceptance criteria:

* The plugin dependency is added to the bundle. Version can be taken from BOM
* There is a demo created which demonstrates how to use the plugin inside Jenkinsfile Runner
  * Demo from https://docs.google.com/presentation/d/1y7YnAdnh5WY59g8oIGTsj8sLQ5KXgoV7uUCBkxcTU88/edit?usp=sharing can be used for it:

```groovy
stage('Read Evergreen YAML') {
    node {
        // Discover core version using Pipeline utility steps
        sh 'wget https://raw.githubusercontent.com/jenkins-infra/evergreen/master/services/essentials.yaml'
        def essentialsYaml = readYaml(file: "essentials.yaml")
        echo "Jenkins Evergreen uses the following Core version: ${essentialsYaml.spec.core.version}"
    }
}
```

rajovictor:
Hi @oleg-nenashev,
I was able to run the above `Jenkinsfile` by adding `pipeline-utility-steps` dependency to the **vanilla-package** `pom.xml`.
I believe, the bom entry is maintained in another repo.
Also, I had to add explicit `org.apache.commons:commons-lang3:3.7` dependency since the `org.apache.maven.plugins.enforcer.RequireUpperBoundDeps` was failing the build otherwise.

Please let me know if those pom changes would be eligible for a PR to solve this issue.

oleg-nenashev:
Hi @rajovictor , thanks for your Interest! This issue was actually solved in #200 which was integrated recently. 

> Also, I had to add explicit org.apache.commons:commons-lang3:3.7 dependency since the org.apache.maven.plugins.enforcer.RequireUpperBoundDeps was failing the build otherwise.

This is something we will need to address later, because the dependency comes from the Jenkins core. It does not appear in #200 because of the dependency inclusion order, but the risk is still there. If yu are interested to create a pull request for it on the top of the current branch, it will be much appreciated

rajovictor:
> Hi @rajovictor , thanks for your Interest! This issue was actually solved in #200 which was integrated recently.
> 
[...]

PR #199, opened 2019-10-14
Add Pipeline Utility Steps to the Vanilla Package

fgakk:
- Extended dependency management in root-pom and vanilla-package pom with pipeline-utility-steps version 1.1.1 
- Extended a declarative pipeline test case for vanilla-package with a yamlReader use cases. However, I cannot get the integration tests running. Let us see what CI has to say ;)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -119,6 +119,13 @@ THE SOFTWARE.
         <artifactId>maven-hpi-plugin</artifactId>
         <version>3.5</version>
       </dependency>
+
+      <dependency>
+        <groupId>org.jenkins-ci.plugins</groupId>
+        <artifactId>pipeline-utility-steps</artifactId>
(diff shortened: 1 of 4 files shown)


fgakk:
I see now, there is another pull request for the same issue, which looks better. @oleg-nenashev you can ignore and close this one. Thanks.

oleg-nenashev:
@fgakk Thanks for your comment. Yes, pull request collisions happen sometimes, especially for GitHub issues when it is not easy to self-assign it for an external contributor (easy in Jenkins JIRA). To avoid that, a good practice is to link a ticket #196 and to make a comment that it is in progress there.

Taking your comment, I will merge #200 and close this one. Note that even closed pull requests count towards Hacktoberfest goals, and they are still much appreciated. Thanks a lot for your contribution!

PR #203, opened 2019-10-17
Add `--job-name`, `--build-number` and `--cause` options

rinckm:
Add additional options to specify
- job name
- build number
- cause
We want to be able to specify meaningful values here so they can be used in some plugins.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,8 @@ The executable of Jenkinsfile Runner allows its invocation with these cli option
  -p (--plugins) FILE         : plugins required to run pipeline. Either a
                                plugins.txt file or a /plugins installation
                                directory. Defaults to plugins.txt.
+ -n (--job-name) VAL         : Name of the job the run belongs to, default to
+                               'job'
  -jv (--jenkins-version) VAL : jenkins version to use (only in case 'warDir' is not
                                specified). Defaults to latest LTS.
  -w (--jenkins-war) FILE     : path to exploded jenkins war directory.
--- a/README.md
+++ b/README.md
@@ -127,6 +127,7 @@ The executable of Jenkinsfile Runner allows its invocation with these cli option
                                directory. Defaults to plugins.txt.
  -n (--job-name) VAL         : Name of the job the run belongs to, default to
                                'job'
+ -c (--cause) VAL            : Cause of the run
  -jv (--jenkins-version) VAL : jenkins version to use (only in case 'warDir' is not
                                specified). Defaults to latest LTS.
  -w (--jenkins-war) FILE     : path to exploded jenkins war directory.
--- a/README.md
+++ b/README.md
@@ -127,6 +127,7 @@ The executable of Jenkinsfile Runner allows its invocation with these cli option
                                directory. Defaults to plugins.txt.
  -n (--job-name) VAL         : Name of the job the run belongs to, default to
                                'job'
+ -b (--build-number) N       : BuildNumber of the run (default: 1)
  -c (--cause) VAL            : Cause of the run
  -jv (--jenkins-version) VAL : jenkins version to use (only in case 'warDir' is not
                                specified). Defaults to latest LTS.
--- a/README.md
+++ b/README.md
@@ -125,10 +125,11 @@ The executable of Jenkinsfile Runner allows its invocation with these cli option
  -p (--plugins) FILE         : plugins required to run pipeline. Either a
                                plugins.txt file or a /plugins installation
                                directory. Defaults to plugins.txt.
- -n (--job-name) VAL         : Name of the job the run belongs to, default to
-                               'job'
- -b (--build-number) N       : BuildNumber of the run (default: 1)
- -c (--cause) VAL            : Cause of the run
+ -n (--job-name) VAL         : Name of the job the run belongs to, defaults to 'job'
+ -b (--build-number) N       : Build number of the run, defaults to 1.
+ -c (--cause) VAL            : A string describing the cause of the run.
+                               It will be attached to the build so that it appears in the
+                               build log and is availabled to plug-ins and pipeline steps.
  -jv (--jenkins-version) VAL : jenkins version to use (only in case 'warDir' is not
                                specified). Defaults to latest LTS.
  -w (--jenkins-war) FILE     : path to exploded jenkins war directory.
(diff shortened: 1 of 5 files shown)


bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
```diff
@@ -69,6 +69,24 @@
...
+    /**
+     * Cause of the Run.
+     */
+    @Option(name = "-c", aliases = { "--cause"}, usage = "Cause of the run")
+    public String cause;
+
+    /**
+     * BuildNumber of the Run.
+     */
+    @Option(name = "-b", aliases = { "--build-number"}, usage = "BuildNumber of the run")
```
oleg-nenashev:
```suggestion
    @Option(name = "-b", aliases = { "--build-number"}, usage = "Build number of the run")
```

payload/src/main/java/io/jenkins/jenkinsfile/runner/Runner.java
```diff
@@ -30,21 +34,32 @@
...
 
-        Action[] workflowActions = bootstrap.workflowParameters != null && bootstrap.workflowParameters.size() > 0 ?
-                new Action[] { new SetJenkinsfileLocation(bootstrap.jenkinsfile, !bootstrap.noSandBox),
-                new ParametersAction(bootstrap.workflowParameters
-                                     .entrySet()
-                                     .stream()
-                                     .map(e -> new StringParameterValue(e.getKey(), e.getValue()))
-                                     .collect(Collectors.toList())) } :
-                new Action[] { new SetJenkinsfileLocation(bootstrap.jenkinsfile, !bootstrap.noSandBox) };
+        List<Action> workflowActionsList = new ArrayList<Action>(3);
```
oleg-nenashev:
```suggestion
        List<Action> workflowActionsList = new ArrayList<>(3);
```

README.md
```diff
@@ -125,6 +125,11 @@ The executable of Jenkinsfile Runner allows its invocation with these cli option
  -p (--plugins) FILE         : plugins required to run pipeline. Either a
                                plugins.txt file or a /plugins installation
                                directory. Defaults to plugins.txt.
+ -n (--job-name) VAL         : Name of the job the run belongs to, defaults to 'job'
+ -b (--build-number) N       : Build number of the run, defaults to 1.
+ -c (--cause) VAL            : A string describing the cause of the run.
+                               It will be attached to the build so that it appears in the
+                               build log and is availabled to plug-ins and pipeline steps.
```
oleg-nenashev:
```suggestion
                               build log and becomes available to plug-ins and pipeline steps.
```

oleg-nenashev (changes requested):
Thanks! Job Name and Build Number sound about right, but I am not sure about `cause`. Introducing a new instance type in the Jenkinsfile Runner would do the job for some edge cases, but I would rather expect common causes to be set, e.g. SCM cause. Having a custom cause type limits options for it to be consumed by plugins. Would be great to have more justification for that before merging

rinckm:
> 
> 
> Thanks! Job Name and Build Number sound about right, but I am not sure about `cause`. Introducing a new instance type in the Jenkinsfile Runner would do the job for some edge cases, but I would rather expect common causes to be set, e.g. SCM cause. Having a custom cause type limits options for it to be consumed by plugins. Would be great to have more justification for that before merging

I had a look to the existing causes but from my point of view they don't really match.
For the SCMTriggerCause we would need to have a polling log file.
If we take UserIdCause we would need to create a user for each execution of the Jenkinsfile Runner to avoid problems.
The RemoteCause comes close but I would only define the note but no host.
We could rename the Cause to NoteCause with 'Started with note %s' as description.
We could also name the Cause FileRunnerCause.
What do you think?

rinckm:
I cannot reproduce the failing check locally. Looks like an infrastructure error to me.

oleg-nenashev:
yes, infra and Jenkinsfile Runner CI are quite unstable.
I agree with having an explicit JFR-specific cause as long as it is not
injected by default

oleg-nenashev (changes requested):
Requesting the change as discussed above, but this is mostly ready to go

rinckm:
This change if fully backward compatible. If no --cause cli option is set the code works as before.

rinckm:
Pipeline

```
node {
        stage('Build') {
              echo currentBuild.getBuildCauses().toString()
            }
    }
```
- --cause 'foo'
```
[{"_class":"io.jenkins.jenkinsfile.runner.FileRunnerCause","shortDescription":"Started by foo"}]
```

- -- cause '' 
```
[]
```

- No cause cli option set
```
[]
```

payload/src/main/java/io/jenkins/jenkinsfile/runner/FileRunnerCause.java
```diff
@@ -0,0 +1,32 @@
+package io.jenkins.jenkinsfile.runner;
+
+/**
+ * @author Matthias Rinck
+ */
+class FileRunnerCause extends hudson.model.Cause {
```
oleg-nenashev:
```suggestion
class JenkinsfileRunnerCause extends hudson.model.Cause {
```

oleg-nenashev (approved):
apart from the class naming, it is ready to go imho.
Class name can be changed later, but it would be great to land it so that we close this story

README.md
```diff
@@ -125,6 +125,11 @@ The executable of Jenkinsfile Runner allows its invocation with these cli option
  -p (--plugins) FILE         : plugins required to run pipeline. Either a
                                plugins.txt file or a /plugins installation
                                directory. Defaults to plugins.txt.
+ -n (--job-name) VAL         : Name of the job the run belongs to, defaults to 'job'
```
oleg-nenashev, line 128: Will create a follow-up task to support folders

payload/src/main/java/io/jenkins/jenkinsfile/runner/FileRunnerCause.java
rinckm: Class name is fixed now.

oleg-nenashev (approved):

rinckm:
Just added 7f4ab782f7ae6a9ab82b5853d29aab23933fcb6a. Is this OK too?

rinckm:
Only fixed the commit message from  7f4ab78 to ced2c76.
I'm ready now.

oleg-nenashev:
Thanks @rinckm! I hope to cut a new release within few days

PR #205, opened 2019-10-17
#197 - Include Jenkins Configuration as Code plugin into the Vanilla Package

### LINKED ISSUE #197: Add JCasC to the Jenkinsfile Runner Vanilla Package
https://github.com/jenkinsci/configuration-as-code-plugin is the base way to configure Jenkins modern instances, and it works well with Jenkinsfile Runner.

Acceptance criteria:

* The plugin dependency is added to the bundle. Version can be taken from BOM
* JCasC directory is exposed as a volume so that users can append their configs
* There is a documentation section JCasC with Jenkinsfile Runner
* There is a demo created which demonstrates how to use the plugin inside Jenkinsfile Runner

FYI @casz @timja

literalplus:
### Ellipsis
This PR adds out-of-the-box support for configuration-as-code to configure the Jenkins instance the Vanilla Docker image runs the Jenkinsfile on. A volume is declared so that users may provide their own configuration file(s).
### Technical overview
This PR adds the JCasC plugin as a dependency to the vanilla module, installing it in the Jenkins running in the Docker image. It also adds configuration (via env) to `Dockerfile` so that the configuration is correctly picked up.
### Remaining challenges
- [x] (1) directory name support of JCasC isn't working on my machine. It tries to open a URL connection to the directory, which (for whatever reason) results in a stream of the file names in that directory separated by spaces. This is not valid YAML and causes the parser to throw. The `file://` URL is necessary because the implementation refuses strings that are not URLs. Seems like an issue in JCasC though. If this is a blocker for this PR, I'll try to reproduce it on a normal Jenkins instance. (Should work according to the docs) - please clarify whether we need that
- [x] (2) the empty `jenkins.yaml` provided in `Dockerfile-dev` (needs to be copies to `Dockerfile` as well) results in a warning from JCasC that it's empty. We might want to fill this with a file containing `jenkins:` or provide a reasonable default - I'll implement the first one hopefully tonight and expand it to a full default if necessary (although what would we want to configure? I saw the JDK is configured in code, so maybe that, but that would break non-Docker runs)
- [x] (3) the demo might need expansion - please provide feedback on whether this demo is sufficient
- [x] (4) docs - I'll add these hopefully tonight

@oleg-nenashev ref: #197

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,14 +25,17 @@ RUN mkdir /app && unzip /jenkinsfile-runner/vanilla-package/target/war/jenkins.w
 
 FROM openjdk:8-jdk
 ENV JENKINS_UC https://updates.jenkins.io
+ENV CASC_JENKINS_CONFIG file:///casc/jenkins.yaml
 USER root
-RUN mkdir -p /app /usr/share/jenkins/ref/plugins
+RUN mkdir -p /app /usr/share/jenkins/ref/plugins /casc
+RUN echo "jenkins: {}" >/casc/jenkins.yaml
 COPY --from=jenkinsfilerunner-build /app/jenkins /app/jenkins
 COPY --from=jenkinsfilerunner-build /jenkinsfile-runner/app/target/appassembler /app
 COPY --from=jenkinsfilerunner-build /jenkinsfile-runner/vanilla-package/target/plugins /usr/share/jenkins/ref/plugins
 COPY jenkinsfile-runner-launcher /app/bin
 
 VOLUME /build
+VOLUME /casc
 
 ENTRYPOINT ["/app/bin/jenkinsfile-runner-launcher", \
             "-w", "/app/jenkins",\
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,17 +25,17 @@ RUN mkdir /app && unzip /jenkinsfile-runner/vanilla-package/target/war/jenkins.w
 
 FROM openjdk:8-jdk
 ENV JENKINS_UC https://updates.jenkins.io
-ENV CASC_JENKINS_CONFIG file:///casc/jenkins.yaml
+ENV CASC_JENKINS_CONFIG /usr/share/jenkins/ref/casc
 USER root
-RUN mkdir -p /app /usr/share/jenkins/ref/plugins /casc
-RUN echo "jenkins: {}" >/casc/jenkins.yaml
+RUN mkdir -p /app /usr/share/jenkins/ref/plugins /usr/share/jenkins/ref/casc
+RUN echo "jenkins: {}" >/usr/share/jenkins/ref/casc/jenkins.yaml
 COPY --from=jenkinsfilerunner-build /app/jenkins /app/jenkins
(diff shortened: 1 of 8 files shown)


Dockerfile-dev
```diff
@@ -3,15 +3,19 @@
 
 FROM openjdk:8-jdk
 ENV JENKINS_UC https://updates.jenkins.io
+ENV CASC_JENKINS_CONFIG file:///casc/jenkins.yaml
```
timja: It would be nice to point this to a directory, so that the configuration can be split into multiple files?

literalplus: I tried that, but CasC read the directory as a space-separated list of the file names, which is invalid YAML. I'll take another look if there's a different way to point it to a directory.

demo/casc/Jenkinsfile
```diff
@@ -0,0 +1,11 @@
+pipeline {
```
oleg-nenashev, line 1: Add a small README which describes how to run the demo?

Dockerfile-dev
oleg-nenashev: I agree that it should be a directory by default. If it does not work, it might be a bug in JCasC.

```suggestion
ENV CASC_JENKINS_CONFIG /casc
```

Dockerfile
```diff
@@ -25,14 +25,17 @@ RUN mkdir /app && unzip /jenkinsfile-runner/vanilla-package/target/war/jenkins.w
 
 FROM openjdk:8-jdk
 ENV JENKINS_UC https://updates.jenkins.io
+ENV CASC_JENKINS_CONFIG file:///casc/jenkins.yaml
```
oleg-nenashev: Why no `/app /usr/share/jenkins/ref/casc` or so to avoid new root dirs?

Dockerfile-dev
timja: The jenkins helm chart relies on the directory behaviour: https://github.com/helm/charts/tree/master/stable/jenkins

literalplus: Updating the JCasC plugin from 0.30 to 0.32 solved the issue. Also it takes normal file names without `file://` protocol now.

This change is in the bom as well, but not yet released, so we'll need to override it until a new version is released (changed here: https://github.com/jenkinsci/bom/commit/64c6b7dd6c51d096833fcd01f04fc2551fa20f11)

Dockerfile
literalplus: The idea was to make the path not so long to type, but seeing that the plugins are there as well, `/usr/share/jenkins/ref/casc` makes more sense - changing this.

Dockerfile-dev
oleg-nenashev: Overriding BOM is fine for now

demo/casc/Jenkinsfile
literalplus: I added a README and included a reference to it in the Docker guide. Let me know what you think :)

oleg-nenashev (approved):
Thanks, looks great!

PR #206, opened 2019-10-18
Add Windows Dockerfile for jenkinsfile-runner

slide:
Severely untested jenkinsfile-runner docker image for windows

I am working on porting the test suite.


oleg-nenashev (approved):
Looks fine, kinda worked on my machine. We are in Beta, and I am ready to merge it.

Ideally we should add extra plugins to the  Windows Vanilla image, e.g. Powershell plugin &Co. But it can be done as a follow-up

PR #207, opened 2019-10-20
Add an interactive CLI mode (--cli argument)

### LINKED ISSUE #178: Add a Jenkins CLI command mode
It would be great if Jenkinsfile Runner supported a mode for calling Jenkins CLI commands against itself. It might be useful to get a list of installed plugins (`list-plugins`) or, for example, to generate a support bundle using https://github.com/jenkinsci/support-core-plugin and so on.

E.g.: 

```
docker run -i --rm -v $(pwd)/Jenkinsfile:/workspace/Jenkinsfile jenkins4eval/jenkinsfile-runner --cli list-plugins
```

Ideally it would be great  to just support interactive mode like Jenkins SSH CLI does.

```
docker run -i --rm -v $(pwd)/Jenkinsfile:/workspace/Jenkinsfile jenkins4eval/jenkinsfile-runner --cli
```

It would require a new engine to be added on the top of Jenkinsfile Engine, but it would be useful

literalplus:
# Ellipsis
As suggested by https://github.com/jenkinsci/jenkinsfile-runner/issues/178, this PR adds a way to access the Jenkins CLI from `jenkinsfile-runner`.

# Technical Overview
To support the new mode of execution, a new `JenkinsLauncher` superclass has been extracted from the existing `JenkinsfileRunnerLauncher`, to abstract Jenkins setup capabilities. A new subclass `CliLauncher` has been added that provides a loop forwarding commands via `CliCommand.clone()`. `App` now has a switch on the new `Bootstrap.cliOnly` flag, and decides which implementation to use.

This flag can be set via `--cli` or `-c`, and additionally via a `FORCE_JENKINS_CLI` environment variable. The latter is mainly intended for use with Docker (otherwise we'd need a custom entrypoint script or a separate `Dockerfile` just for CLI).

# Usage
````bash
$ docker run --rm -it -e FORCE_JENKINS_CLI=true jenkinsfile-runner:dev
Connected to Jenkins!
Type 'help' for a list of available commands, or 'exit' to quit.
 > show-plugins
No such command, try 'help'.
 > list-plugins
ace-editor                         JavaScript GUI Lib: ACE Editor bundle plugin                     1.1
...
workflow-multibranch               Pipeline: Multibranch                                            2.21
 > who-am-i
Authenticated as: anonymous
Authorities:
  anonymous
 > bye
$ app/target/appassembler/bin/jenkinsfile-runner -w /tmp/jenkins -p /tmp/jhome/plugins --cli
...
````

I hope this is somewhat similar to what was meant with "new engine" in #178.

Fixes #178

### DIFF
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -58,6 +58,9 @@ Optionally, if you need special Jenkins configuration, you can mount JCasC YAML 
 [Configuration-as-Code documentation](https://github.com/jenkinsci/configuration-as-code-plugin)
 for available options and the [JCasC demo](demo/casc/README.md) for an example.
 
+To get an interactive Jenkins CLI shell in the container, pass
+`-i -e FORCE_JENKINS_CLI=true` to `docker run` as extra parameters.
+
 ## Debug
 In case you want to debug Jenkinsfile Runner, you need to use the "Vanilla" Docker image built following the steps mentioned in the section above.
 
--- a/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
@@ -70,7 +70,7 @@ public class Bootstrap {
             "Note that the folder specified via --runHome will not be disposed after the run.")
     public File runHome;
 
-    
+
     private static final String DEFAULT_JOBNAME = "job";
 
     /**
@@ -107,6 +107,8 @@ public class Bootstrap {
     @Option(name = "-u", aliases = { "--keep-undefined-parameters"}, usage = "Keep undefined parameters if set")
     public boolean keepUndefinedParameters = false;
 
+    @Option(name = "--cli", usage = "Launch interactive CLI.", forbids = { "-v", "--runWorkspace", "-a", "-ns" })
+    public boolean cliOnly;
 
     public static void main(String[] args) throws Throwable {
         // break for attaching profiler
@@ -141,6 +143,10 @@ public class Bootstrap {
             System.exit(0);
         }
 
+        if (System.getenv("FORCE_JENKINS_CLI") != null) {
+            this.cliOnly = true;
+        }
+
         if (this.version != null && !isVersionSupported()) {
             System.err.printf("Jenkins version [%s] not suported by this jenkinsfile-runner version (requires %s). \n",
                     this.version,
@@ -159,7 +165,7 @@ public class Bootstrap {
         }
 
         if (this.jenkinsfile == null) this.jenkinsfile = new File("Jenkinsfile");
-        if (!this.jenkinsfile.exists()) {
+        if (!this.cliOnly && !this.jenkinsfile.exists()) {
             System.err.println("no Jenkinsfile in current directory.");
             System.exit(-1);
         }
--- a/README.md
+++ b/README.md
@@ -118,9 +118,10 @@ The executable of Jenkinsfile Runner allows its invocation with these cli option
                                Requires Jenkins 2.119 or above
  -a (--arg)                  : Parameters to be passed to workflow job. Use
                                multiple -a switches for multiple params
+ --cli                       : Launch interactive CLI. (default: false)
  -u (--keep-undefined-parameters) : Keep undefined parameters if set, defaults
                                     to false.
- -f (--file) FILE            : Path to Jenkinsfile (or directory containing a
+-f (--file) FILE            : Path to Jenkinsfile (or directory containing a
                                Jenkinsfile) to run, default to ./Jenkinsfile.
  -ns (--no-sandbox)          : Disable workflow job execution within sandbox
                                environment
@@ -181,6 +182,8 @@ Advanced options:
   This way you can access the build metadata in `<jenkinsHome>/jobs/job/builds/1`, like the build.xml, logs, and workflow data, even after the container finished.
 
 * The `-ns` and `-a` options can be specified and passed to the image in the same way as the command line execution.
+
+* You may pass `--cli` to obtain an interactive Jenkins CLI session.
 
 ## Docker build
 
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

File: bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java [hunk 1]
```diff
@@ -83,6 +83,9 @@
     @Option(name = "-h", aliases = { "--help"}, usage = "Prints help information.", help = true, forbids = { "-v", "-w", "-p", "-f", "--runWorkspace" })
     public boolean help;
 
+    @Option(name = "-c", aliases = { "--cli"}, usage = "Launch interactive CLI.", forbids = { "-v", "-jv", "--runWorkspace", "-a", "-ns" })
```
oleg-nenashev: not sure why it has to forbid `-jv`

setup/src/main/java/io/jenkins/jenkinsfile/runner/CliLauncher.java
```diff
@@ -0,0 +1,70 @@
...
+                return;
+            }
+            command.main(findArgsFromLineParts(parts), Locale.getDefault(), System.in, System.out, System.err);
+        } catch (Exception e) {
+            e.printStackTrace();
+            System.err.println("An unexpected error occurred executing this command.");
+        }
+    }
+
+    private List<String> findArgsFromLineParts(String[] commandLineParts) {
```
oleg-nenashev: Could it be generalized in `JenkinsLauncher`

setup/src/main/java/io/jenkins/jenkinsfile/runner/CliLauncher.java
```diff
@@ -0,0 +1,70 @@
...
+import java.io.InputStreamReader;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Locale;
+
+/**
+ * Sets up a Jenkins environment that provides an interactive CLI.
+ */
+public class CliLauncher extends JenkinsLauncher {
```
oleg-nenashev: `CLILauncher` to follow the pattern in Jenkins? No strong opinion

DOCKER.md
```diff
@@ -58,6 +58,9 @@ Optionally, if you need special Jenkins configuration, you can mount JCasC YAML
 [Configuration-as-Code documentation](https://github.com/jenkinsci/configuration-as-code-plugin)
 for available options and the [JCasC demo](demo/casc/README.md) for an example.
 
+To get an interactive Jenkins CLI shell in the container, pass
+`-it -e FORCE_JENKINS_CLI=true` to `docker run` as extra parameters.
```
oleg-nenashev: Why do we need it for Docker?

oleg-nenashev (commented):
Having a bootstrap would be a cleaner approach IMHO. Jenkins official Docker images do that, and we may need it here to provide better compatibility with classic images. E.g. `JAVA_OPTS`, `DEBUG`, `JENKINS_OPTS` have a real use-case in JFR.

But it is fine to do it as a follow-up

DOCKER.md
literalplus: ~~You mean the whole feature? I guess the same arguments apply as for normal mode, for debugging an existing installation, I don't see a reason not to support it. I'm not that familiar with all the use cases here though, so I can remove the docs (and the env var?) if desired~~

Ooooh, you mean `-it`! `i` means to keep stdin open, which we need to actually use the interactive CLI. `t` means to allocate a virtual TTY. This is a standard thing I normally use in this kind of situation, but you're right that it also works without `-t`. Changing.

setup/src/main/java/io/jenkins/jenkinsfile/runner/CliLauncher.java
literalplus: I can move it there, but I don't see how this internal utility method belongs there semantically - the only other existing implementation doesn't do command-line parsing and this use-case seems very specific to this one mode.

literalplus:
Thanks for the review @oleg-nenashev!

With "bootstrap" you mean the Docker image and providing a `docker-entrypoint.sh` script that removes the need to handle the env var in the application itself, right? If so, that makes sense & I can do that. Can you think of anything else such a script should do besides `JENKINS_OPTS` and `DEBUG`? ~~Otherwise, can you please elaborate?~~ 

I just noticed that the integration tests are failing due to an `IllegalAccessError` which sounds...interesting, I'll take a look at it & push the solution together with the two resolved discussions if I find one.

**Update:** The `IllegalAccessError` seems unrelated, also occurs on `master`. The tests also fail there and in other PRs. ~~Haven't yet figured out which of the tests is failing and it is related to my changes (I find the shunit output a little unintuitive to read)~~

**Update:** The failing test is `test_pipeline_execution_hangs` in `tests-negative-scenarios.sh`. I'm not sure why this fails.

**Update:** Failures seem to be random. ~~I first thought it was caused by my earlier PR, but running it a second time with the `master` before that also failed it.~~ Possibly related: `Error response from daemon: conflict: unable to delete 044193f3a355 (cannot be forced) - image is being used by running container 42ad8f3a0ce7`

**Update:** I got the tests to pass locally by re-trying in a loop: `while ! bash ./tests-negative-scenarios.sh; do; echo "try again"; done`: I guess we could re-try them on the CI multiple times as well, but this seems more like a hack than a solution.

oleg-nenashev:
> With "bootstrap" you mean the Docker image and providing a docker-entrypoint.sh script that removes the need to handle the env var in the application itself, right?

Yes

> Update: The failing test is test_pipeline_execution_hangs in tests-negative-scenarios.sh. I'm not sure why this fails.

Just ignore it for now, please. CI is still not very stable. I will test it manually before merging

oleg-nenashev:
Retriggering CI

literalplus:
merge conflicts resolved

literalplus:
In the meantime, another flag with the `-c` shorthand was added (for setting the cause), so I have removed that alias.

oleg-nenashev:
I am not a big fan of shortcuts at all :( Thanks for adjusting the PR!

oleg-nenashev (approved):
I tested it manually, and it works like a charm. Thanks!
I will go ahead and merge it without #213, we can integrate it in a follow-up

PR #177, opened 2019-10-06
Update to the Jenkins 2.222.1 LTS baseline + use Jenkins Core BOM file as a dependency source + update dependencies

oleg-nenashev:
Demonstrates usage of the Jenkins BOM in Jenkinsfile Runner. FTR https://github.com/jenkinsci/jenkins/pull/4150
CC @jtnord @markawm @jglick 

Closes #244

PR #208, opened 2019-10-21
Add check of Minimum supported Jenkins version on startup

### LINKED ISSUE #191: Jenkinsfile Runner should fail immediately when running on an unsupported Jenkins version
`jenkins.version` in pom.xml actually defines the minimum supported Jenkins version. But this version is not really checked anywhere, and JFR run may fail randomly. It would be nice to verify the Jenkins version before staring execution.

Jenkins WAR file includes the Jenkins core version in its properties: https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/jenkins/model/Jenkins.java#L5057-L5066 . The code can be copied from there

It is a follow-up to https://github.com/jenkinsci/jenkinsfile-runner/issues/170

glefloch:
closes #191 by adding a check on version number before doing anything.

### DIFF
--- a/bootstrap/pom.xml
+++ b/bootstrap/pom.xml
@@ -18,5 +18,10 @@
       <groupId>commons-io</groupId>
       <artifactId>commons-io</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci</groupId>
+      <artifactId>version-number</artifactId>
+      <version>1.6</version>
+    </dependency>
   </dependencies>
 </project>
--- a/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
@@ -122,6 +122,13 @@ public class Bootstrap {
             System.exit(0);
         }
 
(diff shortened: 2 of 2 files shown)


oleg-nenashev (changes requested):
Thanks! AFAICT the comparison needs to be changed to "greater or equal", because we define minimum Jenkins version, not the only required one

glefloch:
@oleg-nenashev yes, I wanted to do that, but, can I safely convert the version from `String` to `double` ?

oleg-nenashev:
I suggest using https://github.com/jenkinsci/lib-version-number/blob/master/src/main/java/hudson/util/VersionNumber.java which is a standard way to compare versions in Jenkins

bootstrap/pom.xml
```diff
@@ -18,5 +18,10 @@
       <groupId>commons-io</groupId>
       <artifactId>commons-io</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci</groupId>
+      <artifactId>version-number</artifactId>
+      <version>1.6</version>
```
oleg-nenashev, line 24: Will need to take it from BOM after #177

oleg-nenashev (approved):
Thanks!

PR #209, opened 2019-10-22
Add ParameterDefinitions to WorkflowJob

rinckm:
Create clean parameter definitions in the workflow job to avoid the need to set following JAVA_OPTS:
```
-e "JAVA_OPTS=-Dhudson.model.ParametersAction.keepUndefinedParameters=true"
```
Example:
Pipeline
```
node {
        stage('Build') {
               echo "FOO: '"+params.FOO+"'"
               echo "BAR: '"+params.BAR+"'"
               echo "BAZ: '"+params.BAZ+"'"
            }
    }
```
Parameters
```
 -a FOO= -a BAR="" -a BAZ="xzy"
```
Output of the pipeline
```
[Pipeline] echo
FOO: ''
[Pipeline] echo
BAR: ''
[Pipeline] echo
BAZ: 'xzy'

```

### DIFF
--- a/payload/src/main/java/io/jenkins/jenkinsfile/runner/Runner.java
+++ b/payload/src/main/java/io/jenkins/jenkinsfile/runner/Runner.java
@@ -5,6 +5,9 @@ import hudson.model.Cause;
 import hudson.model.CauseAction;
 import hudson.model.Failure;
 import hudson.model.ParametersAction;
+import hudson.model.ParameterDefinition;
+import hudson.model.ParametersDefinitionProperty;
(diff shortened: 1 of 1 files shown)


oleg-nenashev (changes requested):
This approach might cause issues, because the defined parameters might conflict with parameters declared in the Pipeline script itself.  If there is intent to just ignore that specification, why don't you just set a system property on startup? Or just `--keep-undefined-parameters`?

Needs more discussion IMHO

rinckm:
Replaced by #211

PR #211, opened 2019-10-23
Add a `--keep-undefined-parameters` option so that users can pass parameters without declaring them in job properties

rinckm:
Replace #209


oleg-nenashev (approved):
LGTM

oleg-nenashev:
CI is not healty, this PR looks good. Will test it locally before the release

PR #213, opened 2019-10-25
Docker bootstrap script

literalplus:
## Ellipsis
This MR adds a bootstrap script for Docker, which allows us to switch between Jenkinsfile and CLI mode seamlessly without having to mess with environment variables. It also streamlines a few operations in the Dockerfiles to use environment variables.

This is as suggested by @oleg-nenashev in https://github.com/jenkinsci/jenkinsfile-runner/pull/207#issuecomment-544489650, and inherently requires that #207 (CLI mode) be merged first.

Clean diff without the #207 changes here: https://github.com/xxyy/jenkinsfile-runner/compare/178/cli-mode...xxyy:docker-bootstrap-script?expand=1

### DIFF
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -59,7 +59,11 @@ Optionally, if you need special Jenkins configuration, you can mount JCasC YAML 
 for available options and the [JCasC demo](demo/casc/README.md) for an example.
 
 To get an interactive Jenkins CLI shell in the container, pass
-`-i -e FORCE_JENKINS_CLI=true` to `docker run` as extra parameters.
+`-i` to `docker run` and append `cli` after the image name like this:
+
+```bash
+docker run -i --rm jenkinsfile-runner:my-production-jenkins cli
+```
 
 ## Debug
 In case you want to debug Jenkinsfile Runner, you need to use the "Vanilla" Docker image built following the steps mentioned in the section above.
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,29 +16,34 @@ ENV MAVEN_OPTS=-Dmaven.repo.local=/mavenrepo
 RUN mvn compile dependency:resolve dependency:resolve-plugins
 
 FROM maven as jenkinsfilerunner-build
+
+ARG JENKINS_REF=/usr/share/jenkins/ref
+ARG JENKINS_HOME=/app/jenkins
 ENV MAVEN_OPTS=-Dmaven.repo.local=/mavenrepo
 COPY --from=jenkinsfilerunner-mvncache /mavenrepo /mavenrepo
 ADD . /jenkinsfile-runner
 RUN cd /jenkinsfile-runner && mvn package
-RUN mkdir /app && unzip /jenkinsfile-runner/vanilla-package/target/war/jenkins.war -d /app/jenkins && \
-  rm -rf /app/jenkins/scripts /app/jenkins/jsbundles /app/jenkins/css /app/jenkins/images /app/jenkins/help /app/jenkins/WEB-INF/detached-plugins /app/jenkins/winstone.jar /app/jenkins/WEB-INF/jenkins-cli.jar /app/jenkins/WEB-INF/lib/jna-4.5.2.jar
+RUN mkdir /app && \
+  unzip /jenkinsfile-runner/vanilla-package/target/war/jenkins.war -d ${JENKINS_HOME} && \
+  rm -rf ${JENKINS_HOME}/scripts ${JENKINS_HOME}/jsbundles ${JENKINS_HOME}/css ${JENKINS_HOME}/images ${JENKINS_HOME}/help ${JENKINS_HOME}/WEB-INF/detached-plugins ${JENKINS_HOME}/winstone.jar ${JENKINS_HOME}/WEB-INF/jenkins-cli.jar ${JENKINS_HOME}/WEB-INF/lib/jna-4.5.2.jar
 
 FROM openjdk:8-jdk
-ENV JENKINS_UC https://updates.jenkins.io
-ENV CASC_JENKINS_CONFIG /usr/share/jenkins/ref/casc
+
+ARG JENKINS_REF=/usr/share/jenkins/ref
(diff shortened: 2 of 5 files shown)


oleg-nenashev:
I need to review it carefully. I just expected few params in `/app/bin/jenkinsfile-runner` TBH (missed that this script already exists)

literalplus:
The ` /app/bin/jenkinsfile-runner` script is auto-generated by `appassembler-maven-plugin` and mostly just wraps the Java call. I also noticed a `jenkinsfile-runner-launcher` script in the root dir, but I'm not sure if it's also used for something else, so I left it there. That one just handles the `$DEBUG` environment variable.

Although the new script doesn't do much more than that, just moves a few repeated paths into environment variables so we don't have to duplicate the final values in both the Dockerfile and the script.

I'll resolve the new merge conflicts after the review is in to avoid extra traffic on this PR – the affected files seem to change frequently.

oleg-nenashev:
>  also noticed a jenkinsfile-runner-launcher script in the root dir, but I'm not sure if it's also used for something else, so I left it there. That one just handles the $DEBUG environment variable.

Yes, I forgot it exists. It would be the best place for the new CLI argument so that we do not create more small scripts.

literalplus:
I can move the logic into that script. Do you know if it is used anywhere else that we would need to stay compatible with?

What should it do if the required environment variables (jenkins home, jenkins ref dir) are not defined? Error out like now (breaks existing usages of the script, if any) or only pass the ` --runWorkspace`, ` --file`, `--jenkins-war` and ` --plugins` arguments if the variables are set?

We might be able to save this for `--jenkins-war` and `--plugins` by adding them to `ENTRYPOINT`, but still need special handling for `--file` and `--runWorkspace` because those aren't compatible/useful with `--cli`. (That's also the reason for the special handling there)

I'll probably only get around to doing that end of November

oleg-nenashev:
> I can move the logic into that script. Do you know if it is used anywhere else that we would need to stay compatible with?

The script is used only for Docker images. Jenkinsfile Runner is in Beta, hence there is no strict compatibility requirements ATM

> What should it do if the required environment variables (jenkins home, jenkins ref dir) are not defined?

Not sure I get it. AFAICT there is no serious changes needed being compared with the current behavior

literalplus:
> Not sure I get it.

not an issue if it's only used for Docker :) I thought it might be used for launching in standalone as well, in that case people with existing setups might call that script without the environment variables set

literalplus:
Updated to expand the existing script

oleg-nenashev:
retriggering CI

oleg-nenashev:
Sorry for nt acting on this PR. The Dockerfile patch is still relevant, but I have a slightly different patch in mind: let's modify how CLI works by adding a top-level command. E.g. it would be:

* `jenkinsfile-runner [run] [args...]` for common mode
* `jenkinsfile-runner cli [args...]` for CLI
* `jenkinsfile-runner manage-plugins [args...]` for the plugin manager
*  `jenkinsfile-runner help` for the top-level help
*  `jenkinsfile-runner version` for the version info

It would make the CLI more extensible in the future, and it will help to simplify the Dockerfile as well

literalplus:
That sounds like a good idea! I assume we'd want these commands handled in the app itself instead of the wrapper script then. Thus I'd remove the top-level command handling from the launcher (lines 28-42). However this doesn't make the Dockerfile any simpler than it already is in this PR, so am I understanding this correctly?

I think it makes sense to do refactoring the CLI as a separate task/PR, since it is technically independent from these changes here, which we could then merge for now (minus lines 28-42). I could move the CLI logic into the app for this year's Hacktoberfest if that works for you.

Having top-level commands, we might also be able to split their logic into separate classes, making Bootstrap.java a little slimmer. (I haven't looked at args4j in detail yet, I hope it supports that.)

literalplus:
Sorry for the ping @oleg-nenashev, do you have feedback on what I suggested in my previous comment? Would be great to have this cleared up before Hacktoberfest starts if we want to do it with that :) BTW I'm available to do other stuff as well so if you already have anything specific in mind that needs to be done besides the issues that are already tagged, I'd be happy to help!

re: making it simpler - you're right, we could save the case statement etc. and just always pass -f and --runWorkspace with your suggestion. (= lines 28-42 from above). If the above is fine, I'll rebase this PR (otherwise we could also drop this one and do these changes together with a new top-level command PR)

oleg-nenashev:
Hi @literalplus . My apologies, I missed the notification.

> I assume we'd want these commands handled in the app itself instead of the wrapper script then. Thus I'd remove the top-level command handling from the launcher (lines 28-42). However this doesn't make the Dockerfile any simpler than it already is in this PR, so am I understanding this correctly?

Yes. It is rather a longer-term change when more commands are introduced

> Having top-level commands, we might also be able to split their logic into separate classes, making Bootstrap.java a little slimmer. (I haven't looked at args4j in detail yet, I hope it supports that.)

args4j supports only one-level CLI, so it will require manual implementation. Anyway, it is quite easy to implement multiple option classes with inheritance where needed

> BTW I'm available to do other stuff as well so if you already have anything specific in mind that needs to be done besides the issues that are already tagged, I'd be happy to help!

Any contributions are welcome! Right now I am mostly focused on getting the new packaging flow in place + on performance/size optimizations. But there is a lot of opportunities in the code. https://github.com/jenkinsci/jenkinsfile-runner-image-packs is also an area where many contributions could be made (see CONTRIBUTING)

literalplus:
Since this just popped up in my notifications, a short heads-up: I was unable to participate in this year's Hacktoberfest due to private issues, sorry for not reporting back. I can't yet say when I'll be able to reallocate time for this yet, but will ✨ eventually ✨ be able to take a look (but it's totally fine if it's done by then, there will be other things to do in the future)

oleg-nenashev:
Sure @literalplus . Any contributions are always appreciated and welcome, regardless of Hacktoberfest. I have started implementing https://github.com/jenkinsci/jenkinsfile-runner/issues/429 , it should be done this week. But there is a lot of other stuff that could be done

PR #231, opened 2019-11-20
Add javax.annotation as a dependency

caseycs:
Could not compile project without it locally (osx, oracle-jdk from homebrew)

### DIFF
--- a/bootstrap/pom.xml
+++ b/bootstrap/pom.xml
@@ -23,5 +23,10 @@
       <artifactId>version-number</artifactId>
       <version>1.6</version>
     </dependency>
+    <dependency>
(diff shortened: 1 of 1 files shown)


caseycs:
Turned out it does not work with java11, and with the relevant version it probably compiles correctly. Feel free to decline, if that's the case.

oleg-nenashev:
We are cleaning up the JSR annotations in the core and other components: https://github.com/jenkinsci/jenkins/pull/4604 . I'd guess we will do the same for JFR, and then it won't  be a problem for Java 11 builds (which are neither supported nor tested at the moment)

PR #252, opened 2020-03-08
Add mirror option to speed the download

LinuxSuRen:


bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
```diff
@@ -261,7 +264,7 @@ private File getJenkinsWar() throws IOException {
         File war = new File(cache, String.format("war/%s/jenkins-war-%s.war", version, version));
         if (!war.exists()) {
             war.getParentFile().mkdirs();
-            final URL url = new URL(String.format("http://updates.jenkins.io/download/war/%s/jenkins.war", version));
+            final URL url = new URL(getMirrorURL(String.format("http://updates.jenkins.io/download/war/%s/jenkins.war", version)));
```
LinuxSuRen, line 267: Not sure why we use `http` instead of `https` here.

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
```diff
@@ -47,6 +47,9 @@
     @Option(name = "-p", aliases = { "--plugins" }, usage = "plugins required to run pipeline. Either a plugins.txt file or a /plugins installation directory. Defaults to plugins.txt.")
     public File pluginsDir;
 
+    @Option(name = "-m", aliases = {"--mirror"}, usage = "mirror site of Jenkins, get the mirror list from http://mirrors.jenkins-ci.org/status.html. Defaults to empty.")
```
oleg-nenashev:
```suggestion
    @Option(name = "-m", aliases = {"--mirror"}, usage = "mirror site of Jenkins, defaults to http://updates.jenkins.io/download. Get the mirror list from http://mirrors.jenkins-ci.org/status.html.")
```

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
oleg-nenashev: Our current mirror infrastructure works for HTTP only. For HTTPs you get no mirrors. AFAIK @olblak is exploring replacement options

LinuxSuRen: Got it.

oleg-nenashev: https://github.com/jenkinsci/jenkinsfile-runner/issues/258 as a follow-up

olblak: As @oleg-nenashev mention we don't have a very good support of https at the moment and I am working on different initiatives
There is an open [Pull Request](https://github.com/jenkins-infra/charts/pull/130) for https support of http://archives.jenkins-ci.org
I also deployed [get.jenkins.io](https://get.jenkins.io) which is not as update as mirror.jenkins.io but instead of using mirrorbrain we switch to mirrorbits so we support https on our mirror infrastructure and it also provides more information like [this](https://get.jenkins.io/debian/jenkins_2.224_all.deb?mirrorlist)

Basically the challenges that I am facing now are:
* htacess support, we rely a lot on htacess so I have to switch from nginx to apache or get rid of htacess. I am more considering switch to apache
* Updating artifacts, either we have push or a pull approach
 mirror.azure.jenkins.io has a pull approach and update its data every 5minutes
* https://get.jenkins.io and the futur https://archives.jenkins.io are configured with a push approach but currently it's taking too long, around 15min for each run

olblak: regarding update-center will probably follow once our mirror infra fully support https

olblak: So if you specify https here, you'll just end up on a HTTP mirror at the moment

oleg-nenashev (approved):
I am going to merge it into the master branch. I believe this change is a net improvement

PR #253, opened 2020-03-08
Plugin downloads: Use latest as the default version of plugin

LinuxSuRen:


oleg-nenashev:
To be honest I would not like to introduce the "latest" support at all. But, if it happens, I would rather start from using Plugin Installation Manager tool instead of maintaining yet another implementation in JFR. In such case "latest" support would be introduced anywya.

I am ready to merge as a temporary solution

oleg-nenashev (approved):

PR #262, opened 2020-04-03
Add a Jenkinsfile Runner demo recording to README

oleg-nenashev:
i think it would be great for users to see a simple Pipeline demo on the landing page. Later I plan to switch it to a Vanilla image demo so that everybody can launch the demo in minutes with Docker, but I think we could start from something.

recording


oleg-nenashev:
Will merge so that users see it

PR #265, opened 2020-04-05
Make https default for plugin downloads

### LINKED ISSUE #258: Use HTTPs by default for plugin downloads
Follow-up to https://github.com/jenkinsci/jenkinsfile-runner/pull/252 . Currently Jenkinsfile Runner downloads plugins via HTTP in order to get plugins from mirrors. It would be great to change defaults to HTTPs.

I would expect that runtime plugin downloads are not actively used, but it still worth fixing it.

CC @LinuxSuRen

sladyn98:
This PR aims to make https default for downloading plugins.
Closes: https://github.com/jenkinsci/jenkinsfile-runner/issues/258


sladyn98:
The mirrors still do not support https so I am unsure if it does not find would it default to http 
CC @oleg-nenashev @LinuxSuRen

LinuxSuRen:
I guess we still need to wait until HTTPS is ok.

oleg-nenashev:
Yes, I suggest putting it on hold until the new Mirror infra is rolled out. CC @olblak 
FTR https://issues.jenkins-ci.org/browse/INFRA-2516

rsvoboda:
@oleg-nenashev / @sladyn98 I think this can be rebased and `on-hold` label removed, INFRA-2516 is resolved

oleg-nenashev:
Agreed @rsvoboda

oleg-nenashev:
It worth mentioning that this code will be removed once [Plugin Installation Manager Tool](https://github.com/jenkinsci/plugin-installation-manager-tool) is integrated in #324. I am not a big fan of using on-demand plugin download even with that tool, in the most of cases it is better package plugins in the Build Time while packaging custom JFR images. Good for some demos though

olblak (approved):
\o/ No more HTTP

oleg-nenashev (approved):
Works on my machine (c). I added test coverage in #391 , but this is not a right time to spend too much time on tests for this feature. Tests will be a must when integrating Plugin Installation Manager

PR #267, opened 2020-04-05
Add support for compressed WAR in the `-w` option

### LINKED ISSUE #224: Allow passing a compressed WAR file in the '--war' argument
Currently `--war` option takes a part to the unzipped WAR file directory. Although it is good from the performance point of view, it causes confusion since Jenkinsfile Runner shows an odd behavior (classloading failures) when a WAR file is passed.

I suggest the following behavior:

* If a WAR file is passed, Jenkinsfile Runner automatically unpacks it to a temporary directory OR to a work directory.
** In the case of the work directory (sharing between runs), a checksum should be calculated and persisted somehow to avoid collisions when a source war is changed. Such caching is not mandatory IMHO
* When the option points to a WAR instead of the folder, a warning is printed to indicate the performance implications (WAR unpacking)

CC @btk3000

sladyn98:
This PR aims to allow the jenkinsfile-runner to automatically explode compressed war files.
Closes : #224


bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
```diff
@@ -307,6 +308,15 @@ public int run() throws Throwable {
             return ((IApp) c.newInstance()).run(this);
         }
 
+        // Explode war if necessary
+        if(FilenameUtils.getExtension(warDir.getAbsolutePath()).equals("war")) {
```
oleg-nenashev: Also check that the path is a file?

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
```diff
@@ -307,6 +308,15 @@ public int run() throws Throwable {
...
+            System.out.println("Exploding jenkins.war, this might take some time.");
```
oleg-nenashev: It does not have to be a Jenkins WAR, would be better to use absolute path here.

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
```diff
@@ -307,6 +308,15 @@ public int run() throws Throwable {
...
+            String warPath = warDir.getAbsolutePath();
+            warDir = Util.explodeWar(warPath);
```
oleg-nenashev, line 315: would be better to use another variable so that CLI arguments remain immutable

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Util.java
```diff
@@ -0,0 +1,49 @@
...
+        Enumeration<java.util.jar.JarEntry> enu= jarfile.entries();
+
+        //Creating working directory
+        String currentDirectory = System.getProperty("user.dir");
+        File destDir = new File(currentDirectory + "/work");
+        //Creating the directory
+        if(destDir.mkdir()) {
+            System.out.println("Work directory has been created successfully at" + destDir.getAbsolutePath());
+        } else {
+            System.out.println("Failed to created directory");
```
oleg-nenashev: throw exception?

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Util.java
```diff
@@ -0,0 +1,49 @@
...
+            File file = new File(destDir, je.getName());
+            if(!file.exists()) {
+                file.getParentFile().mkdirs();
+                file = new File(destDir, je.getName());
+            }
+            if(je.isDirectory()) {
+                continue;
+            }
+            InputStream is = jarfile.getInputStream(je);
+            FileOutputStream fo = new FileOutputStream(file);
```
oleg-nenashev: This is not a resource-safe approach. Consider using try-with-resources

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Util.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.Enumeration;
+import java.util.jar.JarEntry;
+import java.util.jar.JarFile;
+
+
+public class Util {
+
+    public static File explodeWar(String jarPath) throws IOException {
+        JarFile jarfile = new java.util.jar.JarFile(new java.io.File(jarPath));
+        Enumeration<java.util.jar.JarEntry> enu= jarfile.entries();
```
oleg-nenashev:
```suggestion
        Enumeration<java.util.jar.JarEntry> enu = jarfile.entries();
```

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Util.java
```diff
@@ -0,0 +1,49 @@
...
+
+        //Creating working directory
```
oleg-nenashev: Since you unpack the WAR every time anyway, it would be better to use a temporary directory. Also, using a generic `work` directory might be a problem for other cases, because "work" in Jenkinsfile Runner is likely to represent a workspace instead

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Util.java
```diff
@@ -0,0 +1,49 @@
...
+        if(destDir.mkdir()) {
+            System.out.println("Work directory has been created successfully at" + destDir.getAbsolutePath());
+        } else {
+            System.out.println("Failed to created directory");
+        }
+
+        while(enu.hasMoreElements()) {
+            JarEntry je = enu.nextElement();
+            File file = new File(destDir, je.getName());
+            if(!file.exists()) {
```
oleg-nenashev: If the directory exists, it is likely to cause dangerous collisions (new files from new WAR, old files from the old one)

oleg-nenashev (changes requested):
Requesting changes because of the "work directory" handling. The approach is very unreliable for those who are running Jenkinsfile Runner not inside a single-shot instance

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
sladyn98: Added :+1

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
sladyn98: Agreed

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
sladyn98: But `.collectJars(new File(warDir,"WEB-INF/lib"))` would need to pick up the WEB-INF file from the exploded war right, so we might need to change this new exploded file path global so it can be used by all functions or pass it as a parameter

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Util.java
sladyn98: Yeah sounds just about right

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Util.java
sladyn98: :+1:

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Util.java
sladyn98: Yeah created a temporary directory which in linux systems would create something like this
`/tmp/java-5122075410590166027`.  So I created a temp dir in the current working directory.

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Util.java
sladyn98: We could delete the temp directory after the class loader has finished loading the necessary classes from the exploded war thus removing this dangerous collision.

sladyn98:
CC @LinuxSuRen

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
LinuxSuRen: What's the problem without exploding the war file?

sladyn98: > Currently --war option takes a part to the unzipped WAR file directory. Although it is good from the performance point of view, it causes confusion since Jenkinsfile Runner shows an odd behavior (classloading failures) when a WAR file is passed.

Refer: https://github.com/jenkinsci/jenkinsfile-runner/issues/224#issue-518539244

oleg-nenashev:
Sorry for missing it. I will review and hopefully integrate it once CI is back to normal

oleg-nenashev:
#285 should bring CI back to normal

oleg-nenashev:
@sladyn98 could you please merge it with the current origin's master?

oleg-nenashev (approved):
This pull request is definitely right, thanks @sladyn98 !
I might need to make some follow-up adjustments for standalone JAR where the Jar is actually not required. But it is a separate story

PR #268, opened 2020-04-05
Add CODEOWNERS to automate review requests

oleg-nenashev:
subj. CC @jenkinsci/jenkinsfile-runner-developers


oleg-nenashev:
I will proceed as is.

PR #281, opened 2020-05-05
[JENKINS-57441] - Fix warning about missing User config path on startup

oleg-nenashev:
The recent Jetty update in #177 led to the repetition of https://issues.jenkins-ci.org/browse/JENKINS-57441 warnings on newer core baselines. This change replaces the Jetty login service by a mock implementation.

It should not only fix the issue but also improve performance. This approach will likely also impact attempts to connect to Jenkinsfile Runner over user/password authentication, but token authentication should continue working correctly.


oleg-nenashev:
I will merge it into the master branch so that it is built with the actual version when I fix tests

PR #288, opened 2020-05-10
Add separate Dockerfile for Maven Cache

oleg-nenashev:
I am moving Maven Cache to a separate CD flow so that it speeds up local and CI builds for production vanilla images


oleg-nenashev:
CI will fail anyway, so merging

PR #298, opened 2020-05-13
Replace jsr-305 annotations with SpotBugs ones

### LINKED ISSUE #296: Replace jsr-305 annotations by SpotBugs ones
Similarly to https://github.com/jenkinsci/jenkins/pull/4604 , we should remove JSR annotations and use ones supplied by SpotBugs.

It should also help towards building JFR with JDK 11, see https://github.com/jenkinsci/jenkinsfile-runner/pull/231

devkonarev:
Fixes #296 

Similarly to jenkinsci/jenkins#4604 , we should remove JSR annotations and use ones supplied by SpotBugs.

It should also help towards building JFR with JDK 11.

This change as it stands introduces several "deprecated usage" compile warnings because of the bom enforced version of `org.jenkins-ci:version-number` library. Once the bom is updated and the library version is changed from `1.6` to at least `1.7`, the warnings will disappear.

### DIFF
--- a/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
@@ -6,7 +6,7 @@ import org.kohsuke.args4j.CmdLineException;
 import org.kohsuke.args4j.CmdLineParser;
 import org.kohsuke.args4j.Option;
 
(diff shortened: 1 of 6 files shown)


oleg-nenashev (approved):
Thanks, it should be enough for the time being.
In the future we should also port the pom.xml patch from https://github.com/jenkinsci/jenkins/pull/4604 , but it can be done https://github.com/jenkinsci/pom so that it applies to all Jenkins library repos

PR #340, opened 2020-07-25
Fix wrong syntax in the documentation

PierreBtz:
This is pretty minor, I guess the remaining of a Makefile :)


oleg-nenashev (approved):
Thanks!

Issue #226, opened 2019-11-06
Jenkinsfile Runner App ZIP and Fat JAR include WARs and plugin JARs while they should not 


oleg-nenashev:
Current design of Jenkinsfile Runner presumes that the plugins are passed using a directory (`-p` option, even if they are mandatory for the JFR well-being). The idea is that the plugin JARs should not be included, but apparently they are bundled in recent JFR versions. It might related to recent Maven Assembly Plugin

![image](https://user-images.githubusercontent.com/3000480/68313590-df1d0600-00c5-11ea-9e50-8626fd93eb88.png)

oleg-nenashev:
I am working on it. Likely I will create two bundles: one without any plugins and the Jenkins WAR, and another portable JAR which includes them

oleg-nenashev:
#449 fixed that. Now we have "slim" packaging which does not include WARs and JARs

PR #347, opened 2020-07-29
Stop bunding plugins in the `app` module + Add new `app-portable` module with the original behavior

oleg-nenashev:
This is an experimental pull request, which stops actual bundling of `payload-dependencies` as it was originally designed by Kohsuke. It removes around 40Mb from the bundles and from the Docker images.

TODO:

- [x] - Modify the dependency tree so that `payload-dependencies` are not included in the JFR dependency tree
- [x] - Create a new portable module which will keep including the Jenkins core and the minimum set of plugins. It could be merged with the Vanilla package later to keep the things simple
- [ ] - Productize classloading, dependency on XStream is not a good idea for a final implementation
- [ ] - Update user and developer documentation

Contributes to #226


oleg-nenashev:
So far it ends up with StackOverflow due to https://github.com/jenkinsci/xstream/blob/master/xstream/src/java/com/thoughtworks/xstream/core/util/CompositeClassLoader.java#L146 . In the current state CompositeClassLoader cannot be a context classloader on its own...

```
java.lang.StackOverflowError
        at sun.net.www.protocol.file.Handler.parseURL(Handler.java:67)
        at java.net.URL.<init>(URL.java:622)
        at java.net.URL.<init>(URL.java:490)
        at sun.misc.URLClassPath$FileLoader.getResource(URLClassPath.java:1311)
        at sun.misc.URLClassPath.getResource(URLClassPath.java:249)
        at java.net.URLClassLoader$1.run(URLClassLoader.java:366)
        at java.net.URLClassLoader$1.run(URLClassLoader.java:363)
        at java.security.AccessController.doPrivileged(Native Method)
        at java.net.URLClassLoader.findClass(URLClassLoader.java:362)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
        at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:349)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:411)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
        at com.thoughtworks.xstream.core.util.CompositeClassLoader.loadClass(CompositeClassLoader.java:136)
        at com.thoughtworks.xstream.core.util.CompositeClassLoader.loadClass(CompositeClassLoader.java:146)
        at com.thoughtworks.xstream.core.util.CompositeClassLoader.loadClass(CompositeClassLoader.java:146)
        at com.thoughtworks.xstream.core.util.CompositeClassLoader.loadClass(CompositeClassLoader.java:146)
        at com.thoughtworks.xstream.core.util.CompositeClassLoader.loadClass(CompositeClassLoader.java:146)
       .....
```

oleg-nenashev:
Will need serious rework after #352 . I am not sure I will be pushing it too much, because packaging to JARs looks a more perspective way for me in longer term. But it would be great to have both

oleg-nenashev:
Done by #449

PR #346, opened 2020-07-29
Exclude jenkins.war and executable-war.jar from Jenkinsfile Runner bundles

oleg-nenashev:
This change removes the `jenkins.war` and `executable-war.jar` resources which were present in the APP packages. It has a massive impact on the bundle size while not having any impact on behavior of Jenkinsfile Runner in all configurations.

* `jenkinsfile-runner-standalone.jar` size was reduced  from 146.5 MB to 82 MB
* `jenkinsfile-runner-standalone.zip` size was reduced from 140 MB  to 75.5 MB
* Vanilla Docker image sizes are reduced by ~70Mb. It is approximately a 20% size reduction for the smallest image.

Contributes to #226 . It worth mentioning that the jenkins-core.jar and plugin JARs remain present in the bundles produced by the `app` module, so there is still some overhead and classloading conflict risks. I plan to submit a separate pull request for it so that these components stop including JARs at all (and maybe adding a separate package for that so that we do not have to pass `--war` and `--plugins` when classpath is ready... at all

PR #348, opened 2020-07-30
Update the Custom WAR Packager demos to use the new plugins and the Jenkins core

oleg-nenashev:
It should address https://github.com/jenkinsci/jenkinsfile-runner/pull/346#issuecomment-665920807 from @PierreBtz


oleg-nenashev:
I will merge it to start assembling the master branch. There are some changes I am working on, and I would like to avoid massive merge conflicts

PierreBtz:
Late to the party but I confirm this fixes the issue raised in #346

Issue #241, opened 2020-01-02
startup speed option


shurikg:
I run the below command and it's take ~11 sec till I see the output from the pipeline.

How can I reduce the start time of jenkinfile-runner ?
```
/users/kube/jenkins-runner -w /users/kube/jenkins -p /users/kube/jenkins_home/plugins -f myJenkinsfile
```

oleg-nenashev:
It is hard to say anything without startup logs and a list of plugin/core versions

shurikg:
where can I find the startup log?

oleg-nenashev:
Sorry, missed the question.
Apparently I also hit the slow startup issue on Windows when using a Vanilla package. With prepared classpath a significant part of the startup time comes from loading extensions. It is not exactly a new discovery, caching the extension list was on the table for JFR for a while.

![image](https://user-images.githubusercontent.com/3000480/81486502-79b4f800-9255-11ea-9ba7-9d2992fcf744.png)

oleg-nenashev:
Curiously, https://github.com/jenkinsci/branch-api-plugin/blob/170e7717413ebaec6e7144fb35344b609fcebe1a/src/main/java/jenkins/branch/CustomOrganizationFolderDescriptor.java#L131-L139 is a major contributor. Likely it happens due to the execution before all extensions are cached.

I could add a High-performance version of ExtensionFinder with caching, but it might be easier to just replace the entire plugin strategy implementation and event the PluginManager.

oleg-nenashev:
I might have a patch submitted in a few days. After extension point optimization I was able to bring down the startup time from 4sec average to around 800ms in the Vanilla image.

shurikg:
thanks for the info, I will try to use the new version.

bennymestel:
Hey @oleg-nenashev was this issue released ? was the run time able to be reduced?

oleg-nenashev:
@bennymestel Right now it is rather an EPIc, not a single change to deliver. I am gradually improving the performance in Jenkinsfile Runner. There is a lot of other things to do, but the recent versions should be considerably better.  If you see any particular issues, please report them separately so that we could tackle them.

PR #358, opened 2020-08-11
Optimize the memory footprint by not keeping stacktraces in AntClassLoader

oleg-nenashev:
These stacktraces are ignored in `ClassicPluginStrategy` anyway... While seemingly minor, this change actually removes more than 200Mb in total heap allocations. I have no access to Intellij IDEA Pro, so no exact metrics. 

* Profiling can be reproduced in https://github.com/jenkinsci/ci.jenkins.io-runner/pull/364

## After

image

## Before (ignore first seconds of slow initialization)

image

Contributes to #241


oleg-nenashev:
Closing in favor of https://github.com/jenkinsci/jenkins/pull/4957 which should achieve the same without any patches on the JFR side. Jenkins core update will be needed

PR #364, opened 2020-09-16
WorkflowJob.resumeBlocked = true

jglick:
There can be no resuming in JFR mode, so do not even pretend.
As a side effect this sets PERFORMANCE_OPTIMIZED as before.


oleg-nenashev (approved):
There might be resuming Pipeline in the future with external Pipeline stage storage, but not now. Thanks for the patch @jglick !

jglick:
Context: [JENKINS-63706](https://issues.jenkins-ci.org/browse/JENKINS-63706)

oleg-nenashev:
@jglick could you please suggest a changelog entry? I can try measuring the performance difference, But I am not sure that anything changes in practice after looking into the code of the current implementations.

jglick:
> I am not sure that anything changes in practice

Not sure if anything does now, but with https://github.com/jenkinsci/workflow-cps-plugin/pull/377 it would.

PR #368, opened 2020-09-19
Allow not specifying any plugin file

PierreBtz:
It's currently not possible to run without specifying any plugin file.


oleg-nenashev (approved):
IIRC from https://github.com/jenkinsci/jenkinsfile-runner/pull/347 , not specifying plugins does not fully work at the moment. Bundled plugins are in classpath and largely work as extension points, but `Jenkins.instance.oluginManager.getPlugin()` returns null for such plugins. It does not block this pull request, but it would definitely benefit from some documentation until the metadata packaging issue is resolved

PierreBtz:
MMhh, I must admit I went with the simplest of pipeline (`echo 'hello world'`) and I can confirm this not very realistic use case works. I'll try to play a bit more and drop a line in the doc.

oleg-nenashev:
@PierreBtz Should not be a problem once the new Packaging flow is finalized. I plan to inject bundled plugin metadata so that the JFR plugin manager properly loads it even if there is no HPIs referenced

PR #369, opened 2020-09-19
Make sure to pass an absolute directory for JENKINS_HOME

PierreBtz:
Not doing this results in an issue when loading Jenkins in https://github.com/jenkinsci/jenkins/blob/9f48a332e5ff542a32d95caeb02b583e5f043829/core/src/main/java/jenkins/model/Jenkins.java#L3204


oleg-nenashev (approved):
LGTM though I have not hit an issue with it yet

PierreBtz:
Reproduction case:
* current head of master
* using 2.249.1
* `runHome` points to a relative directory

PR #370, opened 2020-09-19
Add a withHooks flag to provide a hook directory

### LINKED ISSUE #359: Support passing Groovy Hooks to Jenkinsfile Runner as an argument
Groovy hooks are still popular for doing configurations on startup: https://www.jenkins.io/doc/book/managing/groovy-hook-scripts/ . Currently Jenkinsfile Runner JAR bundles cannot use Groovy Hooks, because it requires either bundling of Groovy Hooks in Jenkins WAR resources or storing them in JENKINS_HOME. Jenkins Home is currently a temporaty directory. 

## Suggested solution

* Option 1: Update the Jenkins core to support passing arbitrary path to Groovy hooks as System property or environment variable
* Option 2: Support passing JENKINS_HOME directory as an argument in Jenkinsfile Runner

## Workarounds

* Option 1: Use JCasC and the https://plugins.jenkins.io/configuration-as-code-groovy/ plugin. Groovy scripts will be executed earlier in the initialization cycle, but it should not matter for the majority of use-cases
* Option 2: Use Custom WAR Packager which injects Groovy files as Jenkins.war resources, and hence Jenkins core is able to pick them up

PierreBtz:
fixes #359

### DIFF
--- a/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
@@ -74,6 +74,10 @@ public class Bootstrap {
             "Note that the folder specified via --runHome will not be disposed after the run.")
     public File runHome;
 
+    @CheckForNull
+    @Option(name = "--withHooks", usage = "Path to a directory containing Groovy Hooks to copy into init.groovy.d")
+    public File withHooks;
+
 
     private static final String DEFAULT_JOBNAME = "job";
 
--- a/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
@@ -75,8 +75,8 @@ public class Bootstrap {
     public File runHome;
 
     @CheckForNull
-    @Option(name = "--withHooks", usage = "Path to a directory containing Groovy Hooks to copy into init.groovy.d")
-    public File withHooks;
+    @Option(name = "--withInitHooks", usage = "Path to a directory containing Groovy Init Hooks to copy into init.groovy.d")
+    public File withInitHooks;
 
 
     private static final String DEFAULT_JOBNAME = "job";
--- a/setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsEmbedder.java
+++ b/setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsEmbedder.java
@@ -331,9 +331,9 @@ public abstract class JenkinsEmbedder implements RootAction {
     }
 
     /**
-     * Prepare the newly allocated home with additional files
+     * Prepare the newly allocated home with additional files.
      * Currently only Groovy Hooks
-     * @param home: the target home directory (freshly allocated)
+     * @param home the target home directory (freshly allocated)
      * @throws IOException any issue during the copy should interrupt the execution
      */
     protected abstract void setupHome(File home) throws IOException;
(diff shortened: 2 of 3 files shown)


PierreBtz:
Tested with and without the `runHome` flag, both are supported

setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsEmbedder.java
```diff
@@ -316,6 +316,8 @@ protected Hudson newHudson() throws Exception {
         jettyLevel(Level.WARNING);
         ServletContext webServer = createWebServer();
         File home = homeLoader.allocate();
+        setupHome(home);
+        // TODO looks like a remaining of Jenkins Test Harness not used at all in the jfr context
```
oleg-nenashev, line 320: Yes. I had a plan to use it for additional extension points inside Jenkinsfile Runner, but I have not got to it yet

setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsEmbedder.java
```diff
@@ -328,6 +330,14 @@ protected Hudson newHudson() throws Exception {
         }
     }
 
+    /**
+     * Prepare the newly allocated home with additional files
+     * Currently only Groovy Hooks
+     * @param home: the target home directory (freshly allocated)
```
oleg-nenashev:
```suggestion
     * @param home the target home directory (freshly allocated)
```

setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsEmbedder.java
```diff
@@ -328,6 +330,14 @@ protected Hudson newHudson() throws Exception {
         }
     }
 
+    /**
+     * Prepare the newly allocated home with additional files
```
oleg-nenashev:
```suggestion
     * Prepare the newly allocated home with additional files.
```

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
```diff
@@ -74,6 +74,10 @@
             "Note that the folder specified via --runHome will not be disposed after the run.")
     public File runHome;
 
+    @CheckForNull
+    @Option(name = "--withHooks", usage = "Path to a directory containing Groovy Hooks to copy into init.groovy.d")
```
oleg-nenashev: Rename to 'withInitHooks'? There are other types of hooks which exist in Jenkins. Boot Failure hooks would be reasonable for some JFR use-cases

oleg-nenashev (approved):
Thanks @PierreBtz !
This approach will work, though it is not very efficient from the performance point of view. Ideally there should be support for custom Groovy hooks sources in the Jenkins core so that JFR could just point a custom directiry

PierreBtz:
@oleg-nenashev another alternative to avoid copying the directory would be to symlink `init.groovy.d` to the provided hook directory. It should work without issue on unix systems, my fear is Windows...
At worst if the fs does not support symlinks we can fall back to the copy.
Wdyt?

oleg-nenashev:
I think it is easier to have a proper patch in the Jenkins core. We could
make Groovy Hooks dir configurable via a system property, and then just set
this property in the Jenkinsfile Runner bootstrap code.

Again, it can be done in a follow-up PR. I will merge it as is if you are
fine with other comments

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
PierreBtz: makes sense

PierreBtz:
@oleg-nenashev suggestions lgtm, I think it's all good now. I created https://issues.jenkins-ci.org/browse/JENKINS-63833 to track the follow up, I might give it a crack if (when? :)) I find the time.

oleg-nenashev (approved):
Looks good to me, thanks!

PR #372, opened 2020-09-21
scala.yml

Eiygy1992:


oleg-nenashev:
This project has no Scala inside

PR #380, opened 2020-10-06
Initial native `checkout scm` support

abayer:
Initial support for `checkout scm` via JCasC-style YAML representations of credentials and SCM.

cc @oleg-nenashev


oleg-nenashev:
@abayer do you have a strong preference about using the XML format? I would rather prefer to avoid adding explicit XStream dependencies in the bootstrap code, and using JCasC YAMLs instead.

abayer:
@oleg-nenashev https://github.com/jenkinsci/jenkinsfile-runner/pull/380/commits/4738114b15339e4f9a9c4c64906767525adc10e6 switches to YAML. =)

tests/pom.xml
```diff
@@ -4,7 +4,7 @@
   <parent>
     <groupId>io.jenkins.jenkinsfile-runner</groupId>
     <artifactId>parent</artifactId>
-    <version>1.0-beta-12-SNAPSHOT</version>
+    <version>1.0-beta-18-SNAPSHOT</version>
```
oleg-nenashev, line 7: 🤷‍♂️

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
```diff
@@ -118,6 +120,12 @@
     @Option(name = "--cli", usage = "Launch interactive CLI.", forbids = { "-v", "--runWorkspace", "-a", "-ns" })
     public boolean cliOnly;
 
+    @Option(name = "--credentials", usage = "YAML credentials definitions to load")
```
oleg-nenashev: Would it make sense to pass SCM and Credentials through the same file? IIUC this file is largely optional, credentials supplied through the main Jenkins.yaml should work as well

payload-dependencies/pom.xml
```diff
@@ -62,6 +62,10 @@
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>scm-api</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.jenkins</groupId>
+      <artifactId>configuration-as-code</artifactId>
```
oleg-nenashev, line 67: It makes JCasC a mandatory requirement for all Jenkinsfile Runner instances. I do not mind.

oleg-nenashev (changes requested):
Looks pretty good, thanks @abayer 👍 ! I am requesting changes to indicate that we need some documentation for this feature (CLI documentation in README + feature documentations somewhere, maybe a new page in https://github.com/jenkinsci/jenkinsfile-runner/tree/master/docs/using).

tests/pom.xml
abayer: Yeah, that was from when I was trying to get the its to work. =)

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
abayer: Lemme think on that.

payload-dependencies/pom.xml
abayer: Yeah, I couldn't come up with a better approach, sadly.

abayer:
@oleg-nenashev Yup, I need to write docs. Will try to do so today.

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
abayer: Yeah, I like that better. Update incoming.

payload/src/main/java/io/jenkins/jenkinsfile/runner/Runner.java
```diff
@@ -53,9 +57,18 @@ public int run(Bootstrap bootstrap) throws Exception {
...
+                Credentials fromSCM = scm.getCredential();
+                if (fromSCM != null) {
+                    addCredentials(fromSCM);
+                }
+                w.setDefinition(new CpsScmFlowDefinition(scm.getSCM(), bootstrap.jenkinsfile.getName()));
+            } else {
+                w.setDefinition(new CpsScmFlowDefinition(
+                        new FileSystemSCM(bootstrap.jenkinsfile.getParent()), bootstrap.jenkinsfile.getName()));
+            }
+            workflowActionsList.add(new SetJenkinsfileLocation(bootstrap.jenkinsfile, !bootstrap.noSandBox));
```
abayer, line 77: So I think in a perfect world, we'd actually just use the Jenkinsfile from the SCM when using an SCM, rather than the one on disk, but that would mean reimplementing [all of this](https://github.com/jenkinsci/workflow-cps-plugin/blob/master/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java#L102-L192), in order to still allow disabling the sandbox. Which is not worth it. So...

payload/src/main/java/io/jenkins/jenkinsfile/runner/Runner.java
```diff
@@ -113,4 +137,13 @@ private void writeLogTo(PrintStream out) throws IOException, InterruptedExceptio
             }
         }
     }
+
+    private void addCredentials(Credentials creds) throws IOException {
```
abayer: This and `getStore()` made sense as separate functions when we were explicitly loading credentials in a loop (or particularly in this case, during the brief moment I was supporting both credentials files and embedded credentials in the scm file) but I feel like keeping them as separate methods still makes sense in case we want to do things like that in the future.

vanilla-package/src/test/java/io/jenkins/jenkinsfile/runner/vanilla/SmokeTest.java
```diff
@@ -130,17 +144,84 @@ public void shouldSupportDataboundMethods() throws Throwable {
...
+            } catch (Exception e) {
+                throw new GitException("unable to write file", e);
+            }
+            git.add(fileName);
+        }
+        git.setAuthor(johnDoe);
+        git.setCommitter(johnDoe);
+        git.commit("Test commit");
+
+        Map<String,Object> remoteConfig = Stream.of(new Object[][]{
```
abayer, line 197: I don't know if there's a better way to do arbitrary map setup than this - the `toYaml` stuff in JCasC is all `NoExternalUse` so I can't just create the `GitSCM` and credentials and then `toYaml` them without being a bad boy. =)

abayer:
@oleg-nenashev Docs added.

README.adoc
```diff
@@ -160,6 +160,10 @@ The executable of Jenkinsfile Runner allows its invocation with these CLI option
  -c (--cause) VAL            : A string describing the cause of the run.
                                It will be attached to the build so that it appears in the
                                build log and becomes available to plug-ins and pipeline steps.
+--scm FILE                   : A YAML file defining the SCM and optional credentials to use
```
oleg-nenashev, line 163: Would be nice to also reference the documentation page from somewhere in README. Can be done as a follow-up pull request

payload-dependencies/pom.xml
oleg-nenashev: Not a problem IMHO. We can keep it mandatory for now, and revisit if it causes issues.

payload/src/main/java/io/jenkins/jenkinsfile/runner/Runner.java
```diff
@@ -113,4 +137,13 @@ private void writeLogTo(PrintStream out) throws IOException, InterruptedExceptio
...
+        CredentialsStore store = getStore();
+        if (store == null) {
+            throw new RuntimeException("Credentials specified but could not find credentials store");
```
oleg-nenashev: Please use checked exceptions and propagate them correctly

payload/src/main/java/io/jenkins/jenkinsfile/runner/Runner.java
```diff
@@ -53,9 +57,18 @@ public int run(Bootstrap bootstrap) throws Exception {
           w.setDefinition(new PipelineAsYamlScriptFlowDefinition(
```
oleg-nenashev, line 58: No `-scm` handling here

oleg-nenashev (changes requested):
Thanks @abayer ! It looks almost ready to go. I am requesting changes because of the two issues I see:

* no option support in the Pipeline as YAML mode. It would be great to implement such support if https://github.com/jenkinsci/pipeline-as-yaml-plugin/issues/28 is not a blocker for that. If not possible, please document the lack of support and throw a exception when Pipeline as YAML is invoked with the flag
* Error propagation via Runtime exceptions

payload/src/main/java/io/jenkins/jenkinsfile/runner/Runner.java
abayer: It's been too long - any suggestions as to what exception would be appropriate here?

abayer:
@oleg-nenashev So it does look like Pipeline as YAML doesn't want to work right - `checkout scm` doesn't work because it's creating a `CpsFlowDefinition`, not a `CpsScmFlowDefinition`. I'll table that.

payload/src/main/java/io/jenkins/jenkinsfile/runner/Runner.java
abayer: Decided to go with `CredentialsUnavailableException`.

payload/src/main/java/io/jenkins/jenkinsfile/runner/Runner.java
oleg-nenashev: Yes. Patch in the upstream may make sense, at least on the API level

oleg-nenashev (approved):
Looks good to me, thanks @abayer ! I might be unable to release it today, but I will try to do it tomorrow on the morning in my TZ

PR #382, opened 2020-10-06
Use platform-specific line breaks in system output operations

### LINKED ISSUE #374: Cleanup SpotBugs warnings in the codebase
There is a number of SporBugs static analysis warnings in the old parts of the Jenkinsfile Runner codebase. It is possible to see the warnings in the build log. Although there is no severe issues, it would be great to cleanup these warnings and then enforce SpotBugs checks during the build.

List of the current SpotBugs violations: https://ci.jenkins.io/job/jenkinsfile-runner/job/jenkinsfile-runner/job/master/lastSuccessfulBuild/analysis/

Acceptance criteria:

* All SpotBugs violations need to be either fixed or suppressed 
* FindBugs failOnError is reenabled in the pom.xml

NOTE: Contributors can submit pull requests for particular issues and issue types, no need to fix everything in a single pull request.

babusekaran:
4 minor spotbugs fixed  for issue type: `[ERROR] Format string should use %n rather than \n`  under https://github.com/jenkinsci/jenkinsfile-runner/issues/374

reffered spotbugs from the build log in https://ci.jenkins.io/job/jenkinsfile-runner/job/jenkinsfile-runner/job/master

### DIFF
--- a/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
@@ -156,7 +156,7 @@ public class Bootstrap {
         }
 
         if (this.version != null && !isVersionSupported()) {
-            System.err.printf("Jenkins version [%s] not suported by this jenkinsfile-runner version (requires %s). \n",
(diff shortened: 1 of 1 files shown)


oleg-nenashev (approved):
Thanks @babusekaran ! This is a valid fix though there are only a few platforms where it might actually lead to problems.

PR #386, opened 2020-10-08
Use adoc code listing instead of md style

rsvoboda:
Use adoc code listing instead of md style


oleg-nenashev (approved):
I agree it's better to use the standard Asciidoc syntax. In practice, it should not be a problem. Modern rendering tools accept the most of GitHub Flavoured Markdown syntax in Asciidoc anyway, but linters might disagree.

oleg-nenashev:
Thanks for your contribution @rsvoboda !

PR #391, opened 2020-10-09
Add Code Coverage Publishing to the build pipeline

oleg-nenashev:
Adds publishing of Code Coverage reports to Jenkins, with summary reporting to GitHub Checks. It is going to be my Hall of Shame for a while. It is not THAT bad, just a lot of dead code in features we do not really need.

image

## Jenkins output

image

## GitHub Checks

image

Could be simplified later if https://github.com/jenkinsci/pom/pull/138 gets merged

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+target/
 app/target/
 bootstrap/target/
 payload-dependencies/target/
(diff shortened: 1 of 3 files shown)


oleg-nenashev:
I will go ahead and merge it. It does not impact the production code

PR #402, opened 2020-10-20
Parent Packaging POM: Allow overriding Jenkinsfile Runner version

oleg-nenashev:
Currently the Jenkinsfile Runner version is hardcoded to the Parent POM version. It leads to two issues:

* It is not possible to create downstream parent POMs with independent versioning
* It is not possible to override JFR versions. In the future I would like to detach Parent POM to another repository, and it may become a use-case for it.

### DIFF
--- a/packaging-parent-pom/pom.xml
+++ b/packaging-parent-pom/pom.xml
@@ -13,6 +13,7 @@
 
   <properties>
     <jfr.name>jenkinsfile-runner-custom-build</jfr.name>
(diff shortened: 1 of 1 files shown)


abayer (approved):

oleg-nenashev:
TL;DR: It is a bad idea when combining this approach with maven-dependency-plugin. I will debug it locally, likely via a series of direct pushes in the release flow

PR #405, opened 2020-10-21
Add support for using custom logger configuration + Add Troubleshooting Guide

oleg-nenashev:
Adds support for overriding log settings  and some documentation about troubleshooting techniques.

CC @michaelneale


michaelneale (approved):
nice touch with the docs too.

PR #411, opened 2020-10-27
Create ClassLoaders inside doPrivileged() + Auto-close the WAR file after exploding it

recrsn:
Partially addresses #374

### DIFF
--- a/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
@@ -18,6 +18,9 @@ import java.net.URL;
 import java.net.URLClassLoader;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
+import java.security.AccessController;
+import java.security.PrivilegedActionException;
+import java.security.PrivilegedExceptionAction;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
@@ -41,7 +44,7 @@ public class Bootstrap {
 
     @Option(name = "-jv", aliases = { "--jenkins-version"}, usage = "jenkins version to use (only in case 'warDir' is not specified). Defaults to latest LTS.")
     public String version;
-    
+
     /**
      * Where to load plugins from?
      */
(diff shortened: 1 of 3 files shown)


bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
```diff
@@ -338,12 +341,12 @@ public int run() throws Throwable {
         }
     }
 
-    public ClassLoader createJenkinsWarClassLoader() throws IOException, NoSuchMethodException, InvocationTargetException, IllegalAccessException {
```
oleg-nenashev, line 341: I do not recall what led to this exception list

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
```diff
@@ -338,12 +341,12 @@ public int run() throws Throwable {
...
-        return new ClassLoaderBuilder(new SideClassLoader(getPlatformClassloader()))
-                .collectJars(new File(warDir,"WEB-INF/lib"))
+    public ClassLoader createJenkinsWarClassLoader() throws PrivilegedActionException {
+        return AccessController.doPrivileged((PrivilegedExceptionAction<ClassLoader>) () -> new ClassLoaderBuilder(new SideClassLoader(getPlatformClassloader()))
```
oleg-nenashev, line 345: Jenkins is unlikely to work on advanced security policies . Should be fine though I will need to add Autotests for the resource loading

oleg-nenashev (approved):
Should be good though I will need to add a test for the AccessController and WAR resource loading

oleg-nenashev (approved):
I tested it locally to verify the workaround for https://github.com/jenkinsci/jenkinsfile-runner/issues/413, and I confirm it still works well after the patch

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
jglick: This is pointless. See https://github.com/spotbugs/spotbugs/issues/1515

PR #417, opened 2020-11-05
Resolved managed Maven plugin dependency versioning

ohbus:
Removed **explicit versioning** to help eliminate conflicts for three **parent managed dependencies**

- [x] org.apache.maven.plugins.**maven-assembly-plugin**

- [x] org.apache.maven.plugins.**maven-resources-plugin**

- [x] org.jvnet.maven-antrun-extended-plugin.**maven-antrun-extended-plugin**

### DIFF
--- a/app/pom.xml
+++ b/app/pom.xml
@@ -41,7 +41,6 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-assembly-plugin</artifactId>
-        <version>3.3.0</version>
         <configuration>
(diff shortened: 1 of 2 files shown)


oleg-nenashev (approved):
I do not recall what exactly led to having these dependencies pinned explicitly. Likely it was a copy-paste at some point. Parent POM now includes recent versions for all Maven plugins it declares, so it should not be a problem to drop explicit version definitions

oleg-nenashev:
Thanks for the patch!

PR #418, opened 2020-11-05
Cleanup SpotBugs warnings in the codebase

ohbus:
Addressing #374 partially, the following are the changes

- [x] Updating exit mechanism

- [x] normalize File paths from using String directly

### DIFF
--- a/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
@@ -154,7 +154,7 @@ public class Bootstrap {
 
         if (showVersion) {
             System.out.println(getVersion());
-            System.exit(0);
+            Runtime.getRuntime().halt(0);
         }
 
         if (System.getenv("FORCE_JENKINS_CLI") != null) {
@@ -175,7 +175,7 @@ public class Bootstrap {
             System.out.println("\nUsage: jenkinsfile-runner [options] [params]\n");
             System.out.println("Options:");
             parser.printUsage(System.out);
-            System.exit(0);
+            Runtime.getRuntime().halt(0);
         }
 
         if (this.jenkinsfile == null) this.jenkinsfile = new File("Jenkinsfile");
--- a/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
@@ -147,7 +147,7 @@ public class Bootstrap {
     /**
      * Directory used as cache for downloaded artifacts.
      */
-    private File cache = new File(FilenameUtils.getName(System.getProperty("user.home") + "/.jenkinsfile-runner/"));
+    private File cache = new File(System.getProperty("user.home") + "/.jenkinsfile-runner/");
 
     @PostConstruct
     public void postConstruct(CmdLineParser parser) throws IOException {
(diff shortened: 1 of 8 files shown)


ohbus:
@oleg-nenashev any thoughts on the failing test cases, or why the build is failing?

I believe this already closed few Static Analysis Reports.

oleg-nenashev (changes requested):
AFAICT `FilenameUtils.getName` does not handle files without extension well (e.g. `Jenkinsfile`). I also do not understand what was the need to apply it everywhere, including logic where folders are legitimately present in the path.

I do not think that `PATH_TRAVERSAL_IN` is actually an issue anywhere in the codebase, because Jenkinsfile Runner deliberately operates on low level with the host filesystem. If a user wants to exploit the filesystem access, it can be done from Pipeline steps.

ohbus:
> AFAICT `FilenameUtils.getName` does not handle files without extension well (e.g. `Jenkinsfile`). I also do not understand what was the need to apply it everywhere, including logic where folders are legitimately present in the path.
> 
> I do not think that `PATH_TRAVERSAL_IN` is actually an issue anywhere in the codebase, because Jenkinsfile Runner deliberately operates on low level with the host filesystem. If a user wants to exploit the filesystem access, it can be done from Pipeline steps.

Let me work on this a little more. Think I stretched it unnecessarily.

ohbus:
**6** Static Analysis Issues has been addressed.

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
```diff
@@ -154,7 +154,7 @@ public void postConstruct(CmdLineParser parser) throws IOException {
 
         if (showVersion) {
             System.out.println(getVersion());
-            System.exit(0);
+            Runtime.getRuntime().halt(0);
```
michaelneale, line 157: this will in theory shut faster - no shutdown hooks right? will it show up as an error in any shells though?

ohbus: This will enable graceful shutdown.

I am not sure about the showing up as an error in any shells though. How can I be sure?

michaelneale: I think it may be the other way around: https://www.baeldung.com/java-runtime-halt-vs-system-exit#:~:text=exit()%20method%20triggers%20the,()%20terminates%20the%20JVM%20abruptly.

if its graceful that you want, then perhaps leave it the way it was?

oleg-nenashev:
Will try to re-review this week

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
```diff
@@ -275,7 +275,7 @@ private File getJenkinsWar() throws IOException {
 
         System.out.printf("Running pipeline on jenkins %s%n", version);
 
-        File war = new File(cache, String.format("war/%s/jenkins-war-%s.war", version, version));
+        File war = new File(cache, FilenameUtils.getName(String.format("war/%s/jenkins-war-%s.war", version, version)));
```
oleg-nenashev, line 278: This one will not work, `war/%s/` will be trimmed by the method

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java
```diff
@@ -288,8 +288,8 @@ private File getJenkinsWar() throws IOException {
 
     private void installPlugin(String shortname, String version) throws IOException {
 
-        final File install = new File(pluginsDir, shortname + ".jpi");
-        File plugin = new File(cache, String.format("plugins/%s/%s-%s.hpi", shortname, shortname, version));
+        final File install = new File(pluginsDir, FilenameUtils.getName(shortname + ".jpi"));
+        File plugin = new File(cache, FilenameUtils.getName(String.format("plugins/%s/%s-%s.hpi", shortname, shortname, version)));
```
oleg-nenashev, line 292: This one will not work, `plugins/%s/` will be trimmed by the method

oleg-nenashev (changes requested):
`FilenameUtils.getName()` might suppressing the issue, but it changes the behavior by removing the path. Please see https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/FilenameUtils.html#getName(java.lang.String) @ohbus 

TBH I do not think it worth spending any time on fixing the path tampering warnings. If there was an issue with that (it is not an API), the user will be able to tamper the files from the Pipeline as well. My recommendation would be to just suppress the warning.

Pretty much the same for `DM_EXIT` warning, the exist is deliberately called in these paths. This warning should be just suppressed IMHO

ohbus:
Great then!

oleg-nenashev:
@ohbus are you still interested in this change? If not, I will close this PR

oleg-nenashev:
@ohbus are you still interested in this change? If not, I will close this PR

ohbus:
I will create a new PR for fixing these changes as there are unnecessary commits to this PR.

Thanks for the reminder @oleg-nenashev

Issue #344, opened 2020-07-28
Create Jenkinsfile Runner Packaging flow for Maven


oleg-nenashev:
Jenkins Vanilla Package has been updated in beta-12 to implement the simple Maven packaging flow which allows packaging plugins and the Jenkins core, verifying dependencies and running functional tests to verify Pipelines. Similar packaging flow could be used by developers who want to customize their Jenkinsfile Runner bundles by using Maven. 

Acceptance criteria:

* Packaging flow is detached from the Vanilla pom.xml into a separate module as a Parent POM. It includes common features like BOM and verification steps so that the bundle verification happens automatically. Packaging should happen to JARs, Docker flow could be added later (in Maven or in Custom WAR Packager).
* Testing utilities are detached to a separate module as a Java lib targeting JUnit 4. Later they could be extended to support JCasC and Groovy hooks supplied by user definitions
* There is documentation about using this flow
* There is a standalone repository demonstrating using the maven packaging flow
 
NOTE: This flow does not compete with https://github.com/jenkinsci/custom-war-packager though it might be able to do packaging faster and better. Later Custom WAR Packager could switch to using the Maven flow under the hood.

oleg-nenashev:
Working on it

oleg-nenashev:
https://github.com/jenkinsci/jenkinsfile-runner-image-packs , still needs documentation

Issue #392, opened 2020-10-09
Parent Packaging POM: Add core version checks for dependencies


oleg-nenashev:
https://github.com/jenkinsci/maven-hpi-plugin/releases/tag/maven-hpi-plugin-3.16 introduced a new `verify-hpi` mojo which checks Core requirements in plugins. It would be great to add core requirement checks to Parent Packaging POM. It should help to prevent core requirement conflicts which are not  detected by the current packaging flow.

PR #430, opened 2020-11-24
Ensure that plugin dependencies are aligned with JFR requirements + Remove the obsolete `databound` demo and test

oleg-nenashev:
Contributes to #344 . After this patch, it will be possible to use custom versions in children packaging POMs.

- ~[ ] Allow using custom version numbers in child packaging POMs~ Reverted in #434 
- ~[ ] Allow setting a Jenkinsfile Runner version via `jfr.version`~ Reverted in #434
- [x] Ensure that plugin dependencies are aligned with JFR requirements
- [x] Remove the obsolete Databound demo


oleg-nenashev:
So this is also "solves" https://github.com/jenkinsci/jenkinsfile-runner/issues/392 for Custom WAR Packager based flows

abayer (approved):
Looks good to me.

Jenkinsfile
```diff
@@ -69,21 +70,7 @@ demos['cwp'] = {
...
-                        sh "make clean buildInDocker run"
-                    }
-                }
-            }
-        }
-    }
-}
+} */
 
 parallel(demos)
```
oleg-nenashev:
```suggestion
}

parallel(demos)
 */

```

oleg-nenashev:
I will merge it to get the boilerplate development going. As documented in https://github.com/jenkinsci/custom-war-packager/issues/193 , it leads to a Custom WAR Packager flow incompatibility, at least for the development versions.

PR #436, opened 2020-11-28
Update Docker builder to Maven 3.6.3, add batch mode flags

oleg-nenashev:
Makes the logs cleaner.

Generally speaking, we need to move the build flow out of DockerHub and to build Jenkinsfile Runner on Jenkins and the Jenkins Release infrastructure (or trusted CI).


PierreBtz (approved):
lgtm. I have a question (unrelated to this PR): would it be worth using a jdk 11 for the builder image of `docker/unix/adoptopenjdk-11-jre-alpine`?

oleg-nenashev:
@PierreBtz we do not officially support building components with JDK11. Many flows work well, for example Jenkins core and plugins. JFR build flow is still unstable. At the same time, it definitely makes sense to use Java 11 target and OpenJDK 11 in the future. It would make the generated bytecode slightly more efficient

PR #447, opened 2020-12-12
Add support for overriding the entrypoint in JFR

oleg-nenashev:
Allows changing the entry point in custom builds. It might be useful for pre-configuration of JFR

### DIFF
--- a/packaging-parent-pom/pom.xml
+++ b/packaging-parent-pom/pom.xml
@@ -188,7 +188,7 @@
               <appendAssemblyId>false</appendAssemblyId>
               <archive>
                 <manifest>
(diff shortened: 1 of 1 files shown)


packaging-parent-pom/pom.xml
```diff
@@ -186,7 +188,7 @@
               <appendAssemblyId>false</appendAssemblyId>
               <archive>
                 <manifest>
-                  <mainClass>io.jenkins.jenkinsfile.runner.bootstrap.Bootstrap</mainClass>
+                  <mainClass>${jfr.mainClass}/mainClass>
```
oleg-nenashev:
```suggestion
                  <mainClass>${jfr.mainClass}</mainClass>
```

PR #449, opened 2020-12-14
Introduce support for size-optimized Jenkinsfile Runner packages (2Mb)

oleg-nenashev:
With this patch, now it is actually possible to package Jenkinsfile Runner separately from the Jenkins core and plugins. The default `app` packaging and vanilla images have been migrated to this approach.

image

Depends on https://github.com/jenkinsci/plugin-installation-manager-tool/pull/243 for a full fix (PIM does not support the desired packaging layout)


oleg-nenashev:
It is working 🥳 Will merge so that we see the new 'latest' bits

PR #450, opened 2020-12-14
Stop publishing the Uber JAR artifact

oleg-nenashev:
For now... It does not do anything good with the slim packaging. For the rest of the packaging it does not work either due to extension indexing issues. To be revised


shadycuz:
@oleg-nenashev This "breaks" your readme. I went to the releases page, download the .zip but it was only the src code. Took me 20 minutes of digging to find this page.

PR #465, opened 2021-01-30
Fix SpotBugs for DM_EXIT warning

ossdhaval:
Suppressed DM_EXIT warning as both conditions rightly require VM to exit


michaelneale (approved):
LGTM - I think this is legit as it is a command line tool.

oleg-nenashev (approved):

oleg-nenashev:
Thanks for the patch! And sorry for the review delay

PR #25, opened 2018-05-17
added brew formula

jamesalbert:
I'm not sure what your plans for release are, but this might make it more easily to get up and running. I was able to install from my repo with the commands added to the readme (replacing "kohsuke" with my name):
```
brew tap kohsuke/jenkinsfile-runner https://github.com/kohsuke/jenkinsfile-runner.git
brew install --HEAD kohsuke/jenkinsfile-runner/jenkinsfile-runner
```

PR #468, opened 2021-02-13
Added brew formula

kennyg:
This is a follow up for https://github.com/jenkinsci/jenkinsfile-runner/pull/25 that fixes the repo names

### DIFF
--- a/Formula/jenkinsfile-runner.rb
+++ b/Formula/jenkinsfile-runner.rb
@@ -1,0 +1,15 @@
+class JenkinsfileRunner < Formula
+    desc "A command line tool to run Jenkinsfile"
+    homepage "https://github.com/jenkinsci/jenkinsfile-runner"
+    head "https://github.com/jenkinsci/jenkinsfile-runner.git"
+
+    bottle :unneeded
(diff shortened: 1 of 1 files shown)


Formula/jenkinsfile-runner.rb
```diff
@@ -0,0 +1,15 @@
+class JenkinsfileRunner < Formula
+    desc "A command line tool to run Jenkinsfile"
```
michaelneale: Change it to "Jenkinsfiles" plural I think - reads a bit nice.

michaelneale (changes requested):
Look good - just pluralise Jenkinsfiles I think

Formula/jenkinsfile-runner.rb
kennyg: Nice catch. I agree.

michaelneale (approved):

oleg-nenashev (approved):
maybe it is better to use GitHub releases instead of rebuilding the tool locally and requiring Maven from HomeBrew. No objection w.r.t merging though, it can be gradually improved

PR #477, opened 2021-03-08
Allow jenkinsfile-runner to integrate with the Jenkins Templating Engine

steven-terrana:
Hello! opening a draft pull request to get some feedback before finalizing this PR. 

### Add JTE options to `PipelineRunOptions`

added a [ `-jte`, `--jenkins-templating-engine`] flag which tells jenkinsfile-runner to use the JTE pipeline definition provider. 

added a [ `-pc`, `--pipeline-configuration` ] flag which allows users to configure the path to a pipeline configuration file for JTE

### Add a JTE Pipeline Definition Provider

Added a `TemplatingEnginePipelineDefinitionProvider` that is able to instrument the `WorkflowJob` with a `FlowDefinition` form JTE. 

### Refactor SCM Credentials onto `SCMContainer`

To avoid code duplication, I refactored the code to add the `-scm` credential to the credential store onto helper functions on the `SCMContainer`.  This allowed the JTE Pipeline Definition Provider to integrate with the remote SCM functionality. 

## Open Questions

* Would it be possible to include this class as part of JTE instead of within the jenkinsfile-runner? Are `ExtensionPoint`s from plugins available to `Runner`?
* If it's possible to load the JTE Pipeline Definition extension point from the JTE plugin itself, then perhaps JTE should also contribute `PipelineRunOptions` by allowing plugins to inject command line options to jenkinsfile-runner. The benefits to this approach would be 3rd party plugins to be able to augment jenkinsfile-runner in isolation without needing to modify jenkinsfile-runner code. Thoughts?
* How would you recommend tests be added for validating JTE? I've done testing myself but i'm unsure how to add tests to a test suite to validate the JTE integration. 

## Checklist

### DIFF
--- a/payload/src/main/java/io/jenkins/jenkinsfile/runner/TemplatingEnginePipelineDefinitionProvider.java
+++ b/payload/src/main/java/io/jenkins/jenkinsfile/runner/TemplatingEnginePipelineDefinitionProvider.java
@@ -1,0 +1,50 @@
+package io.jenkins.jenkinsfile.runner;
+
+import com.cloudbees.plugins.credentials.Credentials;
+import com.cloudbees.plugins.credentials.CredentialsUnavailableException;
+import hudson.Extension;
+import io.jenkins.jenkinsfile.runner.bootstrap.commands.PipelineRunOptions;
+import org.apache.commons.io.FileUtils;
+import org.boozallen.plugins.jte.job.*;
+import org.jenkinsci.plugins.workflow.job.WorkflowJob;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+@Extension public class TemplatingEnginePipelineDefinitionProvider extends PipelineDefinitionProvider {
+
+    private static final Logger LOGGER = Logger.getLogger(TemplatingEnginePipelineDefinitionProvider.class.getName());
+
+    static final String definition = "Jenkins Templating Engine";
+
+    public boolean matches(PipelineRunOptions options) {
+        return options.isJTE;
+    }
+
+    @Override
+    public void instrumentJob(WorkflowJob job, PipelineRunOptions runOptions) throws IOException {
+        AdHocTemplateFlowDefinitionConfiguration config;
+        // check for pipeline configuration
+        File pipelineConfig = runOptions.pipelineConfiguration;
+        if (runOptions.scm != null){
+            SCMContainer scm = SCMContainer.loadFromYAML(runOptions.scm);
+            Credentials creds = scm.getCredential();
+            if(creds != null){
+                scm.addCredentialToStore();
+            }
+            config = new ScmAdHocTemplateFlowDefinitionConfiguration(scm.getSCM(), pipelineConfig.getName(), runOptions.jenkinsfile.getName());
+        } else {
+            String pipelineTemplate = FileUtils.readFileToString(runOptions.jenkinsfile, "UTF-8");
+            boolean hasPipelineConfig = pipelineConfig.exists();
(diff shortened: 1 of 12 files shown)


oleg-nenashev (changes requested):
I am definitely +1 for adding the Jenkins Templating Engine support, but I am not sure about making it a hard requirement. My preference would be to evolve extensibility options so that the plugin becomes optional (see how Pipeline as YAML is implemented now).

Happy to introduce API plugins if needed

steven-terrana:
fully agree, i would prefer to remove `TemplateEnginePipelineDefinitionProvider` from jenkinsfile-runner altogether.  Need to test if the plugin extension point is resolvable from `Runner`. 

Ideally, jenkinsfile-runner shouldn't even need to know about JTE at all. 

I'll see if i can prototype:

1. putting the JTE pipeline definition in the JTE codebase itself rather than in the jenkinsfile-runner source code
2. coming up with an extension point to allow 3rd party plugins to augment jenkinsfile-runner's command line options

If [1] and [2] are possible, then jenkinsfile-runner could integrate without even knowing about JTE

oleg-nenashev:
> coming up with an extension point to allow 3rd party plugins to augment jenkinsfile-runner's command line options

Extension point won't be enough, because plugins and the Jenkins core are not even classloaded at that point in Slim packaging from https://github.com/jenkinsci/jenkinsfile-runner/tree/master/packaging-slim-parent-pom is used. It will require another extension mechanism.

steven-terrana:
set JTE to optional - did a search for `pipeline-as-yaml` and tried to match the dependency style everywhere so they're done the same way.

payload/src/main/java/io/jenkins/jenkinsfile/runner/TemplatingEnginePipelineDefinitionProvider.java
```diff
@@ -0,0 +1,50 @@
...
+import org.apache.commons.io.FileUtils;
+import org.boozallen.plugins.jte.job.*;
+import org.jenkinsci.plugins.workflow.job.WorkflowJob;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+@Extension public class TemplatingEnginePipelineDefinitionProvider extends PipelineDefinitionProvider {
```
oleg-nenashev:
```suggestion
@Extension(optional = true)
public class TemplatingEnginePipelineDefinitionProvider extends PipelineDefinitionProvider {
```

payload/src/main/java/io/jenkins/jenkinsfile/runner/TemplatingEnginePipelineDefinitionProvider.java
```diff
@@ -0,0 +1,50 @@
...
+
+    private static final Logger LOGGER = Logger.getLogger(TemplatingEnginePipelineDefinitionProvider.class.getName());
+
+    static final String definition = "Jenkins Templating Engine";
+
```
oleg-nenashev, line 21: Needs static initializer which checks availability of the class within the plugin.
Otherwise the optional extension will be accepted

oleg-nenashev (changes requested):
Looks good, but the optional exception loading will not work properly with the current implementation

steven-terrana:
hey Oleg - checking in on this. Is there anything else i can do to help get this over the finish line?

oleg-nenashev:
I suppose the ball is on my side, sorry for the delay

oleg-nenashev (approved):
Works on my machine (c). Thanks a lot @steven-terrana ! Will try to release it this week

Issue #210, opened 2019-10-22
Add support of Job Names with folders


oleg-nenashev:
#203 added support of arbitrary Job names, but it is not possible to set a folder using the current approach. It would be great to add support of nested paths like `folder1/folder2/job1` so that users of Jenkinsfile Runner can emulate standard Jenkins structures more easily.

CC @rinckm  @alxsap

glefloch:
Would like to give it a try!

glefloch:
@oleg-nenashev, after looking at the code, it seems like this does not allow for folders.    

    Jenkins.checkGoodName(bootstrap.jobName); 

Would removing this be enough ?

oleg-nenashev:
I would rather create a virtual folder structure inside the JFR runtime.
It is not just enough to have an item with slashes to emulate folders for
some Jenkins plugins

rinckm:
Also environment variable JOB_BASE_NAME should work correctly in the pipeline.

sladyn98:
I was just going through the issue @oleg-nenashev The above PR seems to have closed this issue, is there something pending here ?

oleg-nenashev:
I do not think so @sladyn98 . There are definitely some possible improvements (e.g. for multibranch), but they can be handled separately. Thanks to @addihorn for the patch!

PR #484, opened 2021-03-16
Add support of Job Names with folders (issue #210) + Add JobDSL plugin to the Vanilla Docker image

addihorn:
allows parameter ```--job-name``` to create a job inside a folder-structure.
The parameter will be splitted at ```/``` to allow creation and assignment of jobs to specific folders.

Folders can be created inside ```Runner``` (if not created otherwise) or by Job DSL to allow introduction of additional library-settings etc.

### DIFF
--- a/demo/job-folders/JenkinsfileWithLibraryCall
+++ b/demo/job-folders/JenkinsfileWithLibraryCall
@@ -1,0 +1,14 @@
+
+stage('Read Job Names') {
+	println "JOB_NAME: " + env.JOB_NAME
+	println "JOB_BASE_NAME: " + env.JOB_BASE_NAME
+}
+
+stage('Run Shared Library Code') {
+	println "Return of Library Method infra.isTrusted() : " + infra.isTrusted().toString()
+}
+
+
+
+
+
--- a/demo/job-folders/JenkinsfileWithLibraryCall
+++ b/demo/job-folders/JenkinsfileWithLibraryCall
@@ -7,8 +7,3 @@ stage('Read Job Names') {
 stage('Run Shared Library Code') {
 	println "Return of Library Method infra.isTrusted() : " + infra.isTrusted().toString()
 }
-
-
-
-
-
--- a/demo/job-folders/README.md
+++ b/demo/job-folders/README.md
@@ -1,0 +1,55 @@
+Job Folders demo
+=================
+
+This demo shows, how to define the Job inside of a folder-structure, defined by parameter ```--job-name```  
+To create folders with additional behavior (e.g. loading of shared libraries), Job DSLs as JCasC can be used.
+
+You can run this demos after building jenkinsfile-runner image as described in [Contributing to Jenkinsfile Runner](/CONTRIBUTING.adoc#building)
+
+
+## Run with created Folder by JCasC (with Docker)  
+
+To run in folders without assigned shared libraries  
+
+```bash
+docker run --rm \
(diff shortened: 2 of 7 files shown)


oleg-nenashev (commented):
I am fine with adding Johb DSL to the vanilla package, as long as it is not a mandatory dependency. Actually it is even better since we get support for job definitions in JCasC. Maybe it is another mode to be added later: triggering job from JCasC or Groovy hooks instead of passing Jenkinsfile.  I will create a follow-up ticket, no need to do it here unless you want to do it shortly.

@addihorn before merging, could you please align the codestyle with the existing codebase?

E.g. 

```java
if (condition) {
 // ...
}
```

addihorn:
hi @oleg-nenashev 

Thanks for the review.
Just to understand you correctly
>I am fine with adding Johb DSL to the vanilla package, as long as it is not a mandatory dependency.

You want me, to mark the JobDSL as optional dependency in the vanilla package pom, or is it okay, since the plugin is not referenced in payloads?

oleg-nenashev:
Sorry for confusion. The current version looks good to me function-wise. Just needs some codestyle cleanup. Happy to merge/release after that

addihorn:
Thank you for the clarification. Just wanted to make sure.

Alligned the code with your existing code base, as you asked.

demo/job-folders/README.md
```diff
@@ -40,15 +42,15 @@ docker run --rm \
 
 ## Run named job without folders (with Docker)
 
-```
+```bash
 docker run --rm \
 	-v $(pwd)/demo/job-folders/Jenkinsfile:/workspace/Jenkinsfile \
-	jenkins4eval/jenkinsfile-runner:dev \
+	enkins4eval/jenkinsfile-runner:dev \
```
oleg-nenashev:
```suggestion
	jenkins4eval/jenkinsfile-runner:dev \
```

typo

payload/src/main/java/io/jenkins/jenkinsfile/runner/Runner.java
```diff
@@ -106,6 +125,22 @@ public int run(PipelineRunOptions runOptions) throws Exception {
...
+            Jenkins j = Jenkins.get();
+
+            if(addToFolder==null) {
+                Folder folder = j.getItem(folderName, j, Folder.class);
+                return  folder!=null ? folder : j.createProject(Folder.class, folderName);
+            }
+
+            Folder folder = j.getItem(folderName, addToFolder.getItemGroup(), Folder.class);
+            return folder !=null ? folder : addToFolder.createProject(Folder.class, folderName);
+        } catch (IOException ex) {
```
oleg-nenashev, line 139: Would be nice to add logging there. This is a wrong state which needs to be reported, and it should lead to the execution failure IMHO

demo/job-folders/JenkinsfileWithLibraryCall
```diff
@@ -0,0 +1,14 @@
...
+}
+
+stage('Run Shared Library Code') {
+	println "Return of Library Method infra.isTrusted() : " + infra.isTrusted().toString()
+}
+
+
+
+
+
```
oleg-nenashev:
```suggestion
```

oleg-nenashev (commented):
Would be nice to also address the error propagation before the merge

oleg-nenashev (approved):
Looks good to me. Thanks for addressing the comments @addihorn !

Issue #476, opened 2021-03-07
jenkinsfile-runner-1.0-beta-25 binary Unhandled exception when exploding WAR.


shadycuz:
I get a shell into the  vanilla image:

```
root@ffd1873bfc20:/# bash /app/bin/jenkinsfile-runner-launcher cli
No explicit version has been selected, using latest LTS
Running pipeline on jenkins 2.263.4
Exploding,/root/.jenkinsfile-runner/war/2.263.4/jenkins-war-2.263.4.warthis might take some time.
java.lang.RuntimeException: Unhandled exception
        at io.jenkins.jenkinsfile.runner.bootstrap.commands.JenkinsLauncherCommand.call(JenkinsLauncherCommand.java:59)
        at io.jenkins.jenkinsfile.runner.bootstrap.commands.JenkinsLauncherCommand.call(JenkinsLauncherCommand.java:35)
        at picocli.CommandLine.executeUserObject(CommandLine.java:1933)
        at picocli.CommandLine.access$1200(CommandLine.java:145)
        at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2332)
        at picocli.CommandLine$RunLast.handle(CommandLine.java:2326)
        at picocli.CommandLine$RunLast.handle(CommandLine.java:2291)
        at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2159)
        at picocli.CommandLine.execute(CommandLine.java:2058)
        at io.jenkins.jenkinsfile.runner.bootstrap.Bootstrap.main(Bootstrap.java:45)
Caused by: java.util.zip.ZipException: error in opening zip file
        at java.util.zip.ZipFile.open(Native Method)
        at java.util.zip.ZipFile.<init>(ZipFile.java:225)
        at java.util.zip.ZipFile.<init>(ZipFile.java:155)
        at java.util.jar.JarFile.<init>(JarFile.java:166)
        at java.util.jar.JarFile.<init>(JarFile.java:130)
        at io.jenkins.jenkinsfile.runner.bootstrap.Util.explodeWar(Util.java:60)
        at io.jenkins.jenkinsfile.runner.bootstrap.commands.JenkinsLauncherCommand.postConstruct(JenkinsLauncherCommand.java:91)
        at io.jenkins.jenkinsfile.runner.bootstrap.commands.JenkinsLauncherCommand.call(JenkinsLauncherCommand.java:56)
[...]

PR #493, opened 2021-04-07
Follow redirects when the Jenkins WAR files from update sites

reftel:
- [] Ensure you have provided tests - that demonstrates feature works or fixes the issue

Set the InstanceFollowRedirects flag on connections used to download WAR files etc., since the servers used do use redirects #476


oleg-nenashev (approved):
No objections, it should improve the user experience.
Note that I still do not consider on-demand WAR download a good practice for production deployments

PR #511, opened 2021-05-28
Added a 'lint' command

pawoolley:
Added a `lint` command, which is akin to when a Jenkinsfile is POSTed to `$JENKINS_URL/pipeline-model-converter/validate`.

This has the benefits of being:
1. quicker to lint files because, after the validation stage, if the file passes, the pipeline isn't then executed.
2. safer to lint files because, after the validation stage, the pipeline isn't executed, which might have unwanted effects or change the state of other systems.

Running...
jfr-run

...vs. linting...
jfr-lint

- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -118,6 +118,7 @@ Supported commands:
 
 * `run` - Runs the Jenkinsfile.
    This command also runs by default if no subcommands specified.
+* `lint` - Lints the Jenkinsfile without actually running it.
 * `cli` - Runs interactive https://www.jenkins.io/doc/book/managing/cli/[Jenkins CLI] from where you can access all standard
    Jenkins commands provided by the Jenkins core and installed plugins:
    `list-plugins`, `groovy`, `groovysh`, etc.
--- a/README.adoc
+++ b/README.adoc
@@ -119,6 +119,7 @@ Supported commands:
 * `run` - Runs the Jenkinsfile.
    This command also runs by default if no subcommands specified.
 * `lint` - Lints the Jenkinsfile without actually running it.
+   Only applicable to Declarative Pipeline.
 * `cli` - Runs interactive https://www.jenkins.io/doc/book/managing/cli/[Jenkins CLI] from where you can access all standard
    Jenkins commands provided by the Jenkins core and installed plugins:
    `list-plugins`, `groovy`, `groovysh`, etc.
--- a/setup/src/main/java/io/jenkins/jenkinsfile/runner/App.java
+++ b/setup/src/main/java/io/jenkins/jenkinsfile/runner/App.java
@@ -2,6 +2,8 @@ package io.jenkins.jenkinsfile.runner;
 
 import io.jenkins.jenkinsfile.runner.bootstrap.IApp;
 import io.jenkins.jenkinsfile.runner.bootstrap.commands.JenkinsLauncherCommand;
+import io.jenkins.jenkinsfile.runner.bootstrap.commands.JenkinsfileCommand;
+import io.jenkins.jenkinsfile.runner.bootstrap.commands.LintJenkinsfileCommand;
 import io.jenkins.jenkinsfile.runner.bootstrap.commands.RunJenkinsfileCommand;
 
 /**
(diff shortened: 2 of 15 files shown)


README.adoc
```diff
@@ -118,6 +118,7 @@ Supported commands:
 
 * `run` - Runs the Jenkinsfile.
    This command also runs by default if no subcommands specified.
+* `lint` - Lints the Jenkinsfile without actually running it.
```
jglick, line 121:
```suggestion
* `lint` - Lints the Jenkinsfile without actually running it.
   Only applicable to Declarative Pipeline.
```

vanilla-package/src/test/java/io/jenkins/jenkinsfile/runner/vanilla/SmokeTest.java
```diff
@@ -207,11 +243,17 @@ private String createTestRepoWithContentAndSCMConfigYAML(Map<String,String> file
...
-        git.setAuthor(johnDoe);
-        git.setCommitter(johnDoe);
-        git.commit("Test commit");
+
+        // Commit the staged files.
+        git.commit()
+            .setAuthor(johnDoe)
+            .setCommitter(johnDoe)
+            .setMessage("Test commit")
+            .setSign(false) // Required if you're locally setup for signed commits.
```
pawoolley, line 255: Locally, my gitconfig is setup for signed commits and that was causing this test to fail with a `signing service is not available` exception message.  Using `org.jenkinsci.plugins.gitclient.GitClient`, it was not possible to configure it to disable signed requests.  Switching to `org.eclipse.jgit.api.Git` allows this an lets the test pass for folk with setups like mine.

pawoolley:
> and removed request for jenkinsci/jenkinsfile-runner-developers now

Err...no I didn't 🤔

setup/src/main/java/io/jenkins/jenkinsfile/runner/App.java
```diff
@@ -12,12 +14,21 @@
...
             throw new IllegalStateException(
                     String.format("%s is invoked with a wrong class type. Required=%s, got=%s",
-                            App.class, RunJenkinsfileCommand.class, command.getClass()));
+                            App.class, JenkinsfileCommand.class, command.getClass()));
+        }
+
+        JenkinsLauncher<?> launcher;
+        if (command instanceof RunJenkinsfileCommand) {
+            launcher = new JenkinsfileRunnerLauncher((RunJenkinsfileCommand) command);
+        } else if (command instanceof LintJenkinsfileCommand) {
```
oleg-nenashev, line 26: Later some simplification will be needed for the Command => Launcher mapping. Fine for now

setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsfileLinterLauncher.java
```diff
@@ -0,0 +1,78 @@
...
+    //TODO: add support of timeout
+
+    /**
+     * Launch the Jenkins instance
+     * No time out and no output message
+     */
+    @Override
+    protected int doLaunch() throws Exception {
+        // So that the payload code has all the access to the system
+        ACL.impersonate(ACL.SYSTEM);
```
oleg-nenashev: This is the deprecated call, right? Better to do try-with-resources method so that we do not leak the context

setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsfileLinterLauncher.java
```diff
@@ -0,0 +1,78 @@
...
+
+        Class<?> cc = command.hasClass(CONVERTER_CLASS_NAME) ? Class.forName(CONVERTER_CLASS_NAME) : getConverterClassFromJar();
+        try {
+            // Attempt to call the scriptToPipelineDef method of the Converter class. This is the same as what
+            // happens when a Jenkinsfile is POSTed to $JENKINS_URL/pipeline-model-converter/validate.
+            System.out.println("Linting...");
```
oleg-nenashev: Fine for now though I want to replace everything by loggers later

oleg-nenashev (approved):
Looks good, just a few minor comments. Thanks @pawoolley !

setup/src/main/java/io/jenkins/jenkinsfile/runner/App.java
pawoolley: Agreed.  I did take a quick look at this.  Thought it would be good if each `JenkinsLauncherCommand` had a factory method that allowed it to create the right type of `JenkinsLauncher`, but that would have introduced a circular dependency between the `bootstrap` and `setup` modules.  Will leave it for now while it's still simple.

setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsfileLinterLauncher.java
pawoolley: Can't do a try-with-resources with Java 8, so swapped to a try-finally block.

pawoolley: Wait! What?  Yes, you can do a try-with-resource at Java 8 🤦 
My IDE is lying.  One second, let me check that again.

pawoolley: Ok. Changed it to a try-with-resources.  Also took the liberty of changing `JenkinsfileRunnerLauncher` too, so that both files are consistent.

oleg-nenashev (approved):
CI fails because of the infra issue. The code looks good overall, thanks @pawoolley !

PR #522, opened 2021-06-11
Fails to compile without this

snoopdave:
Without this change I get this error when building via `mvn`

```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  9.314 s
[INFO] Finished at: 2021-06-11T11:07:12-04:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project bootstrap: Compilation failure: Compilation failure:
[ERROR] /Users/dave/src.cloudbees/battleship/jenkinsfile-runner/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/commands/JenkinsLauncherCommand.java:[30,24] cannot find symbol
[ERROR]   symbol:   class PostConstruct
[ERROR]   location: package javax.annotation
[ERROR] /Users/dave/src.cloudbees/battleship/jenkinsfile-runner/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/commands/JenkinsLauncherCommand.java:[79,6] cannot find symbol
[ERROR]   symbol:   class PostConstruct
[ERROR]   location: class io.jenkins.jenkinsfile.runner.bootstrap.commands.JenkinsLauncherCommand
[ERROR] -> [Help 1]
```

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did

Add annotation-api so that `PostConstruct` annotation is defined.

- [x] Link to relevant issues in GitHub or Jira

n/a

- [x] Link to relevant pull requests, esp. upstream and downstream changes

n/a

- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #526, opened 2021-06-15
Removing JSR 250 `@PostConstruct`

jglick:
Rather than #522, simply removing this annotation. I think it was probably unused, seeing as this method was already being explicitly called even in #21, and still is: https://github.com/jenkinsci/jenkinsfile-runner/blob/e76a4461a5f38d02d27d196cb600a6448431867e/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/commands/JenkinsLauncherCommand.java#L67 https://github.com/jenkinsci/jenkinsfile-runner/blob/e76a4461a5f38d02d27d196cb600a6448431867e/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/Bootstrap.java#L61


oleg-nenashev (approved):
Yes, it should be no longer needed after the init chain patches.
Let's merge and see

jglick:
I hope there is test coverage for this?

oleg-nenashev:
> I hope there is test coverage for this?

We are in Beta 🤷

PR #529, opened 2021-06-23
Force to PERFORMANCE_OPTIMIZED for Pipeline execution

abayer:
For some reason, disabling resume isn't doing what I would have expected, i.e., leading to `PERFORMANCE_OPTIMIZED` being used as the durability level. So hey, let's be a little heavy-handed and just set the durability hint for the newly-created job directly.

This also resulted in my IDE cleaning up `Runner.java`'s imports and I decided to allow it.

### DIFF
--- a/payload/src/main/java/io/jenkins/jenkinsfile/runner/Runner.java
+++ b/payload/src/main/java/io/jenkins/jenkinsfile/runner/Runner.java
@@ -1,11 +1,8 @@
 package io.jenkins.jenkinsfile.runner;
 
+import com.cloudbees.hudson.plugins.folder.Folder;
 import com.cloudbees.plugins.credentials.Credentials;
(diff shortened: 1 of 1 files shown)


jglick (approved):

oleg-nenashev (approved):
I'd guess it reverts one of the changes I did last autumn when it was suggested to remove the explicit Durability hint. So I consider it as a bugfix. Thanks for catching it

P.S:  Somebody may have to re-run performance tests

abayer:
@oleg-nenashev Turns out I was confused and reading test case output of a plugin, not JFR output. Blocking resume did do the trick, but I don't think keeping this in would have a downside anyway. It just makes dang sure it uses `PERFORMANCE_OPTIMIZED`. =)

oleg-nenashev:
Yeah. No problem for keeping it explicit. And thanks for checking!

PR #535, opened 2021-07-05
Stop setting `useAntClassLoader` which nothing consumes

basil:
This field has been [unused since 1.527](https://github.com/jenkinsci/jenkins/commit/47de54d070f67af95b4fefb6d006a72bb31a5cb8). Setting it is pointless. Stop doing so, so that one day we might be able to actually remove it.


oleg-nenashev (approved):
Thanks!

PR #537, opened 2021-07-19
Remove `JenkinsEmbedder.contextPath`, default to root

jglick:
The explicit context path `/jenkins` was blindly copied & pasted from https://github.com/jenkinsci/jenkins/commit/1ed4e4b80f73f746f9f896b092ed80005912f500. In the context of JFR it serves no purpose. If for some reason HTTP access to the JFR JVM _is_ being used, we may as well serve it from the root.

Logically this should now be `""` but Jetty inconsistently treats that like `"/"` after issuing a warning: https://github.com/eclipse/jetty.project/blob/34c03ddf1d880ac8c88c21e6530a94b4d03094b0/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java#L1636-L1640

### DIFF
--- a/setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsEmbedder.java
+++ b/setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsEmbedder.java
@@ -120,13 +120,6 @@ public abstract class JenkinsEmbedder implements RootAction {
     protected Server server;
 
     /**
-     * Where in the {@link Server} is Jenkins deployed?
-     * <p>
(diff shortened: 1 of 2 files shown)


agentgonzo (approved):

fcojfernandez (approved):

oleg-nenashev (commented):
I would prefer to see it configurable for compatibility with non-standard Jenkins setup reasons. No problem with changing the defaults

jglick:
@oleg-nenashev any plans for a release? https://github.com/jenkinsci/jenkinsfile-runner/compare/1.0-beta-29...master is getting long.

oleg-nenashev:
No ETA at the moment, not this week unless urgent. Please ping me next week

PR #539, opened 2021-07-20
Make the path that jenkins is accessible on configureable

agentgonzo:
via `--httpPath`
Defaults to `/`

Follow on from #537 allowing for configuration

### DIFF
--- a/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/commands/JenkinsLauncherOptions.java
+++ b/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/commands/JenkinsLauncherOptions.java
@@ -69,6 +69,11 @@ public class JenkinsLauncherOptions {
     public Integer httpPort;
 
     @CheckForNull
+    @CommandLine.Option(names = "--httpPath",
+                        description = "Path (including leading /) for exposing the web server and Jenkins Web UI from Jenkinsfile Runner. Root path (/) by default")
+    public String httpPath;
+
+    @CheckForNull
     @CommandLine.Option(names = "--agentPort",
             description = "Port for connecting inbound Jenkins agents (over JNLP or WebSockets). Disabled by default")
     public Integer agentPort;
--- a/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/commands/JenkinsLauncherOptions.java
+++ b/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/commands/JenkinsLauncherOptions.java
@@ -70,7 +70,7 @@ public class JenkinsLauncherOptions {
 
     @CheckForNull
     @CommandLine.Option(names = "--httpPath",
-                        description = "Path (including leading /) for exposing the web server and Jenkins Web UI from Jenkinsfile Runner. Root path (/) by default")
+                        description = "Path (including leading / but no trailing /) for exposing the web server and Jenkins Web UI from Jenkinsfile Runner. Root path (/) by default")
     public String httpPath;
 
     @CheckForNull
--- a/setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsLauncher.java
+++ b/setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsLauncher.java
@@ -77,7 +77,8 @@ public abstract class JenkinsLauncher<T extends JenkinsLauncherCommand> extends 
                 ? new Server(launcherOptions.httpPort)
                 : new Server(queuedThreadPool);
 
-        WebAppContext context = new WebAppContext(launcherOptions.warDir.getPath(), null);
+        contextPath = launcherOptions.httpPath;
+        WebAppContext context = new WebAppContext(launcherOptions.warDir.getPath(), contextPath);
         context.setClassLoader(getClass().getClassLoader());
         context.setConfigurations(new Configuration[]{new WebXmlConfiguration()});
         context.addBean(new NoListenerConfiguration(context));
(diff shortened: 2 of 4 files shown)


bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/commands/JenkinsLauncherOptions.java
```diff
@@ -68,6 +68,11 @@
             description = "Port for exposing the web server and Jenkins Web UI from Jenkinsfile Runner. Disabled by default")
     public Integer httpPort;
 
+    @CheckForNull
+    @CommandLine.Option(names = "--httpPath",
+                        description = "Path (including leading /) for exposing the web server and Jenkins Web UI from Jenkinsfile Runner. Root path (/) by default")
```
jglick:
```suggestion
                        description = "Path (including leading / but no trailing /) for exposing the web server and Jenkins Web UI from Jenkinsfile Runner. Root path (/) by default")
```

setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsLauncher.java
```diff
@@ -77,7 +77,8 @@ protected ServletContext createWebServer() throws Exception {
                 ? new Server(launcherOptions.httpPort)
                 : new Server(queuedThreadPool);
 
-        WebAppContext context = new WebAppContext(launcherOptions.warDir.getPath(), null);
+        contextPath = launcherOptions.httpPath;
```
jglick: You may need to handle mistakes here:

* trailing `/`
* empty string
* just `/`, which `WebAppContext` will like but `getURL` will not

jglick (commented):
Probably YAGNI but :man_shrugging: @oleg-nenashev requested it.

setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsLauncher.java
agentgonzo: what do you want me to do with all of those. Just throw an Illegal Argument Exception? I agree that probably YAGNI so I don't want to waste too much time with error cases for something that'll probably not get used

jglick: > Just throw an Illegal Argument Exception?

Sure, just to make things more transparent.

> I don't want to waste too much time with error cases for something that'll probably not get used

Agreed.

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/commands/JenkinsLauncherOptions.java
```diff
@@ -68,6 +68,11 @@
             description = "Port for exposing the web server and Jenkins Web UI from Jenkinsfile Runner. Disabled by default")
     public Integer httpPort;
 
+    @CheckForNull
+    @CommandLine.Option(names = "--httpPath",
```
jglick, line 72: FYI for `jenkins.war` the analogous option is

```
   --prefix                 = add this prefix to all URLs (eg http://localhost:8080/prefix/resource). Default is none
```

though that name might be confusingly generic-sounding for something that only affects HTTP service which you would normally not use in JFR to begin with.

setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsLauncher.java
agentgonzo: done. Case 1 and 3 are covered by the same "ends with /" statement

jglick (approved):

fcojfernandez (approved):

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/commands/JenkinsLauncherOptions.java
oleg-nenashev: I prefer the current option. Maybe `--httpPrefix` but no strong opinion

setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsLauncher.java
```diff
@@ -107,6 +109,15 @@ protected ServletContext createWebServer() throws Exception {
         return context.getServletContext();
     }
 
+    private String validateHttpPath(String httpPath) throws IllegalArgumentException {
+        if (StringUtils.endsWith(httpPath, "/")) {
+            throw new IllegalArgumentException("httpPath must not end with a trailing /");
+        } else if ("".equals(httpPath)) {
```
oleg-nenashev, line 115:
```suggestion
        } else if (StringUtils.isBlank(httpPath)) {
```

oleg-nenashev (changes requested):
Documentation would be nice

bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/commands/JenkinsLauncherOptions.java
agentgonzo: I'll leave it as it is then.

setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsLauncher.java
agentgonzo: Not in this case. `null` is a valid input (what you get when the user does not set `--httpPath`) (and `StringUtils.isBlank(null)` returns true)

agentgonzo:
@oleg-nenashev - documentation present now. The build failure seems to be infra-related so can you retrigger a build (or just merge if you are happy as the last commit was only to the readme)

oleg-nenashev (approved):
Sorry for the review delay. Looks good to me!

jglick:
@oleg-nenashev reminder https://github.com/jenkinsci/jenkinsfile-runner/compare/1.0-beta-29...master

PR #541, opened 2021-08-05
Interrupt any running build if the JFR JVM exits cleanly

jglick:
Untested, but should offer better cleanup behavior in case the Pipeline is using any external resources.

### DIFF
--- a/payload/src/main/java/io/jenkins/jenkinsfile/runner/Runner.java
+++ b/payload/src/main/java/io/jenkins/jenkinsfile/runner/Runner.java
@@ -3,9 +3,11 @@ package io.jenkins.jenkinsfile.runner;
 import com.cloudbees.hudson.plugins.folder.Folder;
 import com.cloudbees.plugins.credentials.Credentials;
 import com.cloudbees.plugins.credentials.CredentialsUnavailableException;
+import hudson.init.Terminator;
 import hudson.model.Action;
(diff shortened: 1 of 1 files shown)


oleg-nenashev (commented):
Any particular reason to do that? I am not sure why JFR behavior for Pipeline termination should be different from classic Jenkins

jglick:
> why JFR behavior for Pipeline termination should be different from classic Jenkins

Because in a Jenkins server, exiting should _not_ interrupt running Pipelines. That would obviate all the durability features of Pipeline! Whereas if a JFR container dies for whatever reason while a build was in progress, clearly that build is going to die as well, so you should give it the chance to clean up a bit: send any “aborted” notifications, release locks on external services, whatever.

To be more useful, this should probably wait (up to ~10s) for the build to terminate on its own. The current code sends an abort signal but then allows JVM shutdown to proceed in parallel.

oleg-nenashev (changes requested):
Jenkinsfile Runner also can proceed with Jenkins Pipeline execution when used in the CLI mode with external JENKINS_HOME. While this is not the immediate use-case for the most of the users, I would not like to block it completely. There should be at least a documented opt out flag

jglick:
> proceed with Jenkins Pipeline execution when used in the CLI mode with external JENKINS_HOME

Sorry, what?

PR #591, opened 2021-11-18
Add a simple docker example based on the community docker image

### LINKED ISSUE #557: Make Jenkinsfile Runner accessible for non-Java users
Hey everyone, 

I'm a big fan of the Jenkinsfile-runner, I have been for years when I accidentally stumbled upon it in early development. Unfortunately I have never been able to get it to work. I revisit often and from time to time I see updated instructions and so I try again and again but never am I successful. Well not on my own, see I have actually used the project for a year now but only because I steal working solutions, like the dockerfile from [here](https://github.com/jenkinsci/ci.jenkins.io-runner). 

I'm sure the project is easy to build and run if your very intimate with either Jenkins/Java or both. But I feel this project would be much more successful if you could make it so anyone can run the Jenkinsfile-runner. I think a good first step would be to provide instructions to build a Jenkinsfile-runner based on the [Jenkins](https://github.com/jenkinsci/docker/blob/master/README.md) docker image. Lots of people are already using this image at 500million pulls. This would make it extremely easy to take a companies production Jenkins and then put it in a CI/CD pipeline. Also the steps to configure the `jenkins:jenkins:lts` image is already well defined. So it should make customizing a Jenkinsfile-runner even easier. 

I'm not 100% on the implementation details but I think you could provide two routes.

```Dockerfile
FROM jenkins/jenkins:LTS

// some steps to configure the Jenkins container with plugins etc

// some steps that install the Jenkinsfile-runner
```

and/or

```Dockerfile
FROM jenkins/jenkins:LTS

FROM Jenkinsfile-Runner as runner

COPY --from runner jenkinsfile-runner /usr/bin/

// steps to configure the Jenkins container with plugins etc.
```

That to me would make it so much easier for me to move people to Jenkinsfile-runner or get them started 🚀 

Thanks!

shadycuz:
This PR adds a simple docker example to the `demo` directory and closes #557.

I made the change to the project `README.md` in a separate commit so that is can easily be dropped or edited as the maintainers see fit. I just wanted to add the example, feel free to present it to users how ever you see fit.

I do think it's important to change how JFR is presented to users. There are several ways of running JFR currently.

* You could run JFR locally by downlading/building the binary yourself.
* You could run JFR locally by using the JFR docker image.
* You could run JFR locally be turning the JFR docker image into a replica of the Jenkins you use in production.
* You could run JFR locally by turning your production Jenkins docker image into a JFR.

I don't really think that the first 2 use cases will have lots of users. The third use case seems to be what most of the documentation in the `Extending Docker` section/readme is about and most of the demos. I think this is the reason I struggled to use this tool for so many years, because I was trying to turn my existing Jenkins image into a JFR and not turn the JFR into a copy of my existing Jenkins image. 

Which is why I created this PR. I think the majority of users are wanting to take an existing Jenkins installation and make it run locally via CLI with JFR. I also think this will lower the burden on the JFR maintainers because JFR is actually kinda simple, at least compared to turning the JFR vanilla docker image into an exact replicate of each users Jenkins server. I think most of the issues and examples in the repo are about the Jenkins customization part because the documentation takes you down a path where you are turning the JFR docker image into Jenkins instead of the other way around.

My second point is that if you are going to start pushing the "convert Jenkins to a JFR" methodology. Then I highly recommend you standardize with the community [Jenkins](https://github.com/jenkinsci/docker) image. For example using the same file paths as they do for the war file, Jenkins home and plugins. Which I think can easily be done in #213 and would make it even easier for new users to get started with JFR and Docker. This also means all the customization of Jenkins would be pushed to the community Jenkins repo, which already is thoroughly documented and would allow for the JFR maintainers to worry about jdk versions, jbang etc and less about plugins, groovy init scripts and JCasC.

- [x] ~~Ensure you have provided tests - that demonstrates feature works or fixes the issue~~

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -302,9 +302,9 @@ and then to configure Jenkins so that Pipelines can properly operate in the user
 There are 2 ways to extend Jenkinsfile Runner:
 
 * Using low-level management tools, including the `--plugins` command included in CLI.
-* For Docker images only: using Custom WAR/Docker Packager which automates some build steps and allows managing Jenkinsfile Runner configuration via a single YAML file.
-
-For Docker images see link:./docs/using/EXTENDING_DOCKER.adoc[Extending Jenkinsfile Runner images] for more information and examples.
+* Docker
+** A simple example that converts a Jenkins image into a JFR can be found link:./demo/docker-simple/[here].
+** A more advanced example using the Custom WAR/Docker Packager which automates some build steps and allows managing Jenkinsfile Runner configuration via a single YAML file can be found link:./docs/using/EXTENDING_DOCKER.adoc[here].
 
 == Reporting issues
 
--- a/demo/docker-simple/Dockerfile
+++ b/demo/docker-simple/Dockerfile
@@ -1,0 +1,31 @@
+# I had trouble using jdk8
+ARG baseImage=jenkins/jenkins:latest-jdk11
+
+# Load the JFR docker image
+FROM jenkins/jenkinsfile-runner:latest as jfr
+
+# Lets configure our Jenkins image
+FROM ${baseImage}
+
+# This installs the "Pipeline" plugins
+RUN jenkins-plugin-cli --plugins workflow-aggregator
+
+# Now lets turn our existing docker container into a JFR.
+
+# The jenkins/jenkins containers run as the user "jenkins" which will prevent us from exploding files later.
+USER root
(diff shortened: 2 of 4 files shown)


oleg-nenashev (approved):
Sorry for missing it, and thanks a lot! this is a really good addition

PR #607, opened 2022-02-16
Updating Adding Tools section in EXTENDING_DOCKER.adoc

Ajaypathak372:
This is a simple demo for how to add tools in jenkinsfile-runner. Same can be applied to add other tools.

### DIFF
--- a/docs/using/EXTENDING_DOCKER.adoc
+++ b/docs/using/EXTENDING_DOCKER.adoc
@@ -47,10 +47,13 @@ See the https://github.com/jenkinsci/plugin-installation-manager-tool[Plugin Ins
 
 == Adding tools
 
-To add a tool let's say `git`, you can use https://github.com/jenkinsci/configuration-as-code-plugin[Jenkins Configuration-as-Code].
-First create a Dockerfile to install required plugins for that tool as shown in link:#_installing_plugins[Installing Plugins] section. 
+To add a tool let's say `git`, you can use https://github.com/jenkinsci/configuration-as-code-plugin[Jenkins Configuration-as-Code]. To know more about tools in jenkins, see https://www.jenkins.io/doc/book/managing/tools/[Managing Tools].
 
-Most of the tool plugins in jenkins has support to automatically download tool from internet and install it which can be specified in Jenkins CasC. But If you want to manually install the tool, you can write the installation steps of the tool in the same Dockerfile, like here for `git`.
+To configure a tool in jenkins, you first need to install required plugins for that tool, so you can install plugins in `jenkinsfile-runner` as shown in link:#_installing_plugins[Installing Plugins] section. As `git` plugin is already installed in the `jenkins/jenkinsfile-runner`, so no need to install it again.
+
+*Note:* Here `git` tool is different than `git` plugin, here `git` tool is the `git` command that jenkins internally uses to clone repositories.
+
+Most of the tool plugins in jenkins has support to automatically download tool from internet and install it which can be specified in Jenkins CasC, like shown in this https://github.com/jenkinsci/configuration-as-code-plugin/tree/master/demos/nodejs[example]. But If you want to manually install the tool, you can write the installation steps of the tool in the same Dockerfile (that used to install plugins), like here for `git`, the preferred method is to install `git` command is with OS Package Manager.
 
 Dockerfile::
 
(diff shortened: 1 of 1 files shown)


docs/using/EXTENDING_DOCKER.adoc
```diff
@@ -47,8 +47,58 @@ See the https://github.com/jenkinsci/plugin-installation-manager-tool[Plugin Ins
...
+jenkins.yaml::
+
+[source]
+----
+tool:
+  git:
+    installations:
+      - name: git
+        home: /usr/bin/git
+----
```
krisstern, line 77: I think maybe it would be better to inform the user of the perspective you are coming from throughout, so they could avoid the pitfall wherein they follow your instructions as is and would likely be getting errors due to setup config issues. For example in here, you could add that say for macOS users who would ordinarily install their `git` using Homebrew or `brew`, their `git` path would alternatively or additionally be something like `/opt/homebrew/bin/git`. As is the code would not work for some macOS and Windows users.

docs/using/EXTENDING_DOCKER.adoc
```diff
@@ -47,8 +47,58 @@ See the https://github.com/jenkinsci/plugin-installation-manager-tool[Plugin Ins
...
+
+Now you can run a Jenkinsfile having `git` as a stage like given below
```
krisstern: It would be better to state what is the intention of this Jenkinsfile content, since some contributors will inevitably be newbies. You can for example say something like "The following is to illustrate a 'Hello, World' example by way of a Jenksinsfile...". Otherwise it would take time for the user of this guide to guess the context of this bit of documentation.

docs/using/EXTENDING_DOCKER.adoc
```diff
@@ -47,8 +47,58 @@ See the https://github.com/jenkinsci/plugin-installation-manager-tool[Plugin Ins
...
+        }
+    }
+}
+----
+You can then run `jenkinsfile-runner` like this
+[source]
+----
+docker run --rm -v $(pwd)/Jenkinsfile:/workspace/Jenkinsfile -v $(pwd)/jenkins.yaml:/usr/share/jenkins/ref/casc/jenkins.yaml jenkinsfile-runner:my-image
+----
+Here `jenkinsfile-runner:my-image` is custom docker image created from above Dockerfile.
```
krisstern, line 102: Instead of concluding with this line, you could extend to other cases by generalising or providing a list of the names of the other tools (as we have discussed on Gitter before), which they could try using the same procedure. Otherwise it appears like this is only useful for the `git` tool alone. With more details the user of the documentation would not feel lost as the stud may appear incomplete to some who would like to learn more. It would indeed be a very nice documentation if the desired effect could be described in detail, perhaps with screenshots to capture the intended outcome.

docs/using/EXTENDING_DOCKER.adoc
```diff
@@ -47,8 +47,58 @@ See the https://github.com/jenkinsci/plugin-installation-manager-tool[Plugin Ins
...
+        stage('Print hello') {
+            tools {git 'git'} //optional
+            steps {
+                git branch: 'master', url: 'git-repo-url'
+            }
+        }
+    }
+}
+----
+You can then run `jenkinsfile-runner` like this
```
krisstern: Better use more formal language like "You can then run `jenkinsfile-runner` like the following:" or "You can then run `jenkinsfile-runner` like what is shown below:" then using the more colloquial phrase "like this" as in casual conversation. Using a colon would definitely help with the punctuation.

docs/using/EXTENDING_DOCKER.adoc
Ajaypathak372: Here we are installing git in jenkinsfile-runner docker image which is linux based, so I think there is no need to specify the installation paths for macOS or windows users.

krisstern: You will have to explain at least something like "the jenkinsfile-runner docker image is linux-based", or people will not be aware of it.

Ajaypathak372: I think the one who knows jenkins, can easily understand this, and also isn't it obvious that it is Linux based docker image for anyone using jenkinsfile-runner docker image?

krisstern: Well we have to consider new-comers to Jenkins as well like yourself.

Ajaypathak372: yeah, you are right but see all docker images created from base images like ubuntu, centos and these are linux based and since here this is the Docker doc for jenkinsfile-runner, so I guess there is no need to mention anything like here.

docs/using/EXTENDING_DOCKER.adoc
Ajaypathak372: Any update about this PR?

oleg-nenashev:
Sorry I definitely missed the last update

oleg-nenashev (commented):
I will need to reword it before merging. Installing tools from the internet should be discouraged in Jenkinsfile Runner. Tools should be baked into the images whenever possible

PR #645, opened 2022-08-10
Setup CD for multiple Docker images on GitHub Actions

oleg-nenashev:
Once completed, there should be proper tags deployed by GitHub actions


oleg-nenashev:
GitHub Actions run is fine. I will merge it to see how the deployment works

PR #740, opened 2026-06-30
fix: restore argLine in vanilla-package surefire config

romanisb:
@oleg-nenashev this should fix the two CI tests failing in https://github.com/jenkinsci/jenkinsfile-runner/runs/84234113409.

### Submitter checklist
- [ ] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

### DIFF
--- a/vanilla-package/pom.xml
+++ b/vanilla-package/pom.xml
@@ -229,6 +229,7 @@
               <excludes>
                 <exclude>**/its/**.java</exclude>
               </excludes>
+              <argLine>-Djdk.net.URLClassPath.disableClassPathURLCheck=true --add-opens=java.base/java.util=ALL-UNNAMED --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.lang.invoke=ALL-UNNAMED ${argLine}</argLine>
             </configuration>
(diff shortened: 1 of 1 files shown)


vanilla-package/pom.xml
```diff
@@ -229,6 +229,7 @@
               <excludes>
                 <exclude>**/its/**.java</exclude>
               </excludes>
+              <argLine>-Djdk.net.URLClassPath.disableClassPathURLCheck=true --add-opens=java.base/java.util=ALL-UNNAMED --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.lang.invoke=ALL-UNNAMED ${argLine}</argLine>
```
romanisb, line 232: The `override`, defined in lin 228, removes the surefire settings from https://github.com/jenkinsci/jenkinsfile-runner/blob/main/pom.xml#L201. Not sure which direction is prefered here. Just reused the configuration again here. Could be considered to clean up all the configuration in the future at one point.

romanisb:
@oleg-nenashev if everything is fine, would it be possible to trigger a new release?

oleg-nenashev (approved):

